PR #4, opened 2019-05-15
Domain

pyotrritchie:


src/Domain/Claim.cs
```diff
@@ -0,0 +1,12 @@
+namespace Linn.Authorisation.Domain
+{
+    public class Claim
```
richardiphillips: I'm not sure I'd use the term Claim in authorisation. We're using it to create System.Whatever.Claim in the auth world at the moment but might not always be the case?

src/Domain/Roles.cs
```diff
@@ -0,0 +1,15 @@
...
+
+    public class Role
+    {
+        public int Id { get; set; }
+
+        public string Name { get; set; }
+
+        public IEnumerable<Claim> Claims { get; set; }
+
+        public IEnumerable<string> Members { get; set; }
```
richardiphillips: Are all user uris attached directly to a role? I think we should be able to group them from the start

richardiphillips: Would the role itself also be returned as a claim? I'm guessing the thought behind having multiple claims on a role is to support specific things like edit.root.product.phase.out or whatever but if we're still planning to let the individual solutions do a lot of this work then we'd just need purchasing.planner or export or whatever.

src/Service.Host/CustomBoostrapper.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Linn.Authorisation.Service.Host
 {
-    using System;
+  /*  using System;
```
richardiphillips: delete?

pyotrritchie: 👍

src/Domain/Roles.cs
pyotrritchie: > Are all user uris attached directly to a role? I think we should be able to group them from the start

Only those users who have that role are attached to it.. The reason Group isn't there is because we thought it was more of a conceptual convenience for when the user is interacting with a form. A quick way of attributing roles to a user. Initially the idea was to initialise the database on first load with the small set of roles that we have right now. And then later, as a part of other work, to add Group and User, and write some reactjs forms to manage them

pyotrritchie: > Would the role itself also be returned as a claim? I'm guessing the thought behind having multiple claims on a role is to support specific things like edit.root.product.phase.out or whatever but if we're still planning to let the individual solutions do a lot of this work then we'd just need purchasing.planner or export or whatever.

Thinking was we would just return the claims of the client and hide the concept of roles from actual services..

src/Domain/Claim.cs
pyotrritchie: Can rename to Permissions?

pyotrritchie:
> I just wonder if we need a domain view from the user object down too?
> When there are hundreds of roles and claims and groups it might be a bit slow to search for a user through them all? Having said that you're probably right not to pre-optimise

I think for the current dataset we can probably get away with this but yeh that's definitely something that may be needed

src/Domain/Roles.cs
andrewfraser73: The design doesn't preclude groups in any way indeed the member property can point to a group.  If we were to implement just the permissions required for products-maint as they were just now you wouldn't need groups.  Groups are a convenience for when you don't want to keep adding the same group of users to a role.

PR #5, opened 2019-05-15
Claim is now Permission

andrewfraser73:
simple rename of Claim to Permission.  just like the Change UK party we like to change our names


src/Service.Host/client/src/helpers/userManager.js
```diff
@@ -12,7 +12,7 @@ var oidcConfig = {
     post_logout_redirect_uri: `${host}`,
     silent_redirect_uri: `${host}/authorisation/signin-oidc-silent`,
     automaticSilentRenew: true,
-    filterProtocolClaims: true,
+    filterProtocolPermissions: true,
```
richardiphillips: replace all victim?

andrewfraser73: looks like it

PR #7, opened 2019-05-21
Refactor

lewisrenfrew:
- misnamed branch, originally started writing the service but switched to refactoring facade and wanted to pr before proceeding
 
- moved logic out of facade into domain services

- should I move facade tests to domain tests?


src/Facade/GroupFacadeService.cs
```diff
@@ -0,0 +1,25 @@
+namespace Linn.Authorisation.Facade
+{
+    using System.Collections.Generic;
+    using Domain.Groups;
+
+    using Linn.Authorisation.Domain.Services;
+    using Linn.Common.Facade;
+
+    public class GroupFacadeService : IGroupFacadeService
```
andrewfraser73: is this used anywhere anymore?  the group service functionality is now in authorisationservice in the domain which uses a group repo?

lewisrenfrew: good spot... not currently in use although it might resurface when we add other group related methods that need to be wrapped in an IResult. Will remove for now either way

PR #8, opened 2019-05-23
Nancy service

lewisrenfrew:
nancy routes to achieve:

- getPrivilegesForMemeber(who)
- CreatePrivilege()
- CreatePermission()

Should we add any more at this point? updates? deletes? Group stuff?


src/Service/Modules/AuthorisationModule.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public sealed class AuthorisationModule : NancyModule
+    {
+        private readonly IAuthorisationService authorisationService;
+
+        public AuthorisationModule(IAuthorisationService authorisationService)
+        {
+            this.authorisationService = authorisationService;
+
+            this.Get("/privileges/{who*}", parameters => this.GetPrivilegesForMember(parameters.who));
```
atcampbell: This might be better as a query string rather than a route, as 'who' will be a uri. 

Something like app.linn.co.uk/authorisation/privileges?who=/users/33067 rather than app.linn.co.uk/authorisation/privileges/users/33067

lewisrenfrew: ooooh yeah, totally forgot we already discussed this 👍

src/Facade/ResourceBuilders/PrivilegeResourceBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+                           Active = privilege.Active,
+                           Links = this.BuildLinks(privilege).ToArray()
+                       };
+        }
+        
+        object IResourceBuilder<Privilege>.Build(Privilege p) => this.Build(p);
+
+        public string GetLocation(Privilege model)
+        {
+            return $"/authorisation/privileges/{model.Name}";
```
andrewfraser73: is there going to be a privilege id which would make the uri /privileges/{privilege.Id}?  if you add name to the uri then any funny characters in the privilege name could cause trouble in the future

src/Facade/PermissionService.cs
```diff
@@ -0,0 +1,95 @@
...
+                return new BadRequestResult<Permission>();
+            }
+
+            var result = this.Add(resource);
+            if (result is BadRequestResult<Permission>)
+            {
+                return result;
+            }
+
+            var privilege = this.privilegeRepository.FilterBy(p => p.Name == resource.Privilege).FirstOrDefault();
```
andrewfraser73: what happens if someone tries to add a privilege that doesn't exist the logic seems like it will make a permission with a null privilege.

src/Facade/PermissionService.cs
```diff
@@ -0,0 +1,95 @@
...
+            var group = this.groupRepository.FindAll().FirstOrDefault(g => g.Name == resource.GroupName);
```
andrewfraser73: also what happens if there is no group with groupName

src/Facade/PermissionService.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            if (resource.GranteeUri != null && resource.GroupName == null)
+            {
+                return new CreatedResult<Permission>(new IndividualPermission(resource.GranteeUri, privilege, resource.GrantedByUri));
```
andrewfraser73: will this get refactored to add it to the repo first when there is a wired up repo? same with group one

PR #9, opened 2019-05-23
Ef model

pyotrritchie:


src/Domain/Groups/GroupMember.cs
```diff
@@ -13,11 +12,16 @@ public GroupMember(Group group, string addedByUri)
             this.DateAdded = DateTime.UtcNow;
         }
 
+        public GroupMember()
```
richardiphillips: I know it's merged already but worth mentioning that all the EF constructors can be private.

PR #10, opened 2019-05-29
Persistence Repos

andrewfraser73:
This adds the repo wiring up to allow privileges and individual permissions to be made
it also changes the route for privilege who lookup to be /authorisation/privileges which matches auth-oidc

### DIFF
--- a/src/Domain/Services/PrivilegeService.cs
+++ b/src/Domain/Services/PrivilegeService.cs
@@ -20,6 +20,8 @@
 
         public IEnumerable<Privilege> GetPrivileges(string who)
         {
+            // TODO what if who is string.Empty ?
+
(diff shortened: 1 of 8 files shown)


src/Domain/Services/PrivilegeService.cs
```diff
@@ -20,6 +20,8 @@ public PrivilegeService(IRepository<Group, int> groupRepository, IRepository<Per
 
         public IEnumerable<Privilege> GetPrivileges(string who)
         {
+            // TODO what if who is string.Empty ?
+
```
lewisrenfrew: I'd deal with this case in the service layer (e.g. with a  request validator or something).. so could move that todo to there

PR #14, opened 2019-06-26
Create and view privileges

iaincrawf:


src/Service.Host/client/src/components/Root.js
```diff
@@ -1,34 +1,69 @@
...
+
+                            <Route
+                                path="/"
+                                render={() => {
+                                    document.title = 'Authorisation';
+                                    return false;
+                                }}
+                            />
+
+                            <Switch store={store}>
```
atcampbell: I don't think Switch needs the store here

src/Service.Host/client/src/containers/Privilege.js
```diff
@@ -0,0 +1,39 @@
...
+const getId = ownProps => ownProps.match.params.id;
+
+const mapStateToProps = (state, { match }) => ({
+    id: match.params.id,
+    privilege: getPrivilege(state)
+});
+
+const mapDispatchToProps = (dispatch, ownProps) => {
+    return {
+        initialise: () => dispatch(fetchPrivilege(getId(ownProps))),
```
atcampbell: I think you could just use the id parameter you've declared in mapStateToProps. `dispatch(fetchPrivilege(id))`

src/Service.Host/client/src/components/App.js
```diff
@@ -1,12 +1,34 @@
...
+
+const styles = () => ({
+    root: {
+        margin: '40px',
+        padding: '40px'
     }
-}
+});
+
+const App = ({ classes }) => (
```
atcampbell: I think its better to use a function declaration for Components, as this way loses the function name and makes debugging more difficult 

`function App({ classes }) { //blah } `

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const handleUpdatePrivilegeName = (propertyName, newValue) => {
+        updatePrivilegeName(newValue);
+    };
+    const handleTogglePrivilegeStatus = () => {
+        togglePrivilegeStatus();
+    };
+
+    const privilegename = privilege.name;
+    const privilegeactive = privilege.active || false;
```
atcampbell: camel case would make these a bit clearer

atcampbell: There's also an argument for putting these in selectors and passing them in from the container. This way they can be tested and updated if the shape of the object changes.

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -0,0 +1,117 @@
...
+    return (
+        <div>
+            <Link to="../viewprivileges">
+                <Button type="button" variant="outlined">
+                    Back to all privileges
+                </Button>
+            </Link>
+            <Paper className={classes.root}>
+                <Title text="View/Edit Privilege" />
+                {privilege ? (
```
atcampbell: It might be better to check explicitly if the privilege is loading here. That would allow you to differentiate between a loading privilege and an error / missing privilege etc

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -0,0 +1,134 @@
...
+    createPrivilege,
+    updateNewPrivilege,
+    privileges,
+    newprivilege
+}) => {
+    useEffect(() => {
+        initialise();
+    }, [initialise]);
+
+    let initialName = '';
```
atcampbell: Do you need this variable if you're just pulling it from the prop? In the table there could be something like `newPrivilege.name || ''`

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -0,0 +1,134 @@
...
+                            <TableCell>Privilege</TableCell>
+                            <TableCell align="right">Active status</TableCell>
+                        </TableRow>
+                    </TableHead>
+                    <TableBody>
+                        {privileges.map(privilege => (
+                            <TableRow
+                                key={privilege.name}
+                                component={Link}
+                                to={privilege.links[0].href.slice(1)}
```
atcampbell: I think this might be better as a selector. We also have a getSelfHref function which will pull out this uri

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -0,0 +1,134 @@
...
+    if (newprivilege) {
+        initialName = newprivilege.name;
+    }
+
+    const handleNameChange = e => {
+        updateNewPrivilege(e.target.value);
+    };
+    const dispatchcreatePrivilege = () => createPrivilege(initialName);
+
+    return privileges ? (
```
atcampbell: Again it would be nice to check if privileges are loading explicitly.

src/Service.Host/client/src/components/Root.js
atcampbell: I don't think you need to pass store into Switch here

src/Service.Host/client/src/containers/Privilege.js
atcampbell: I think you can just use the id parameter you've declared in mapStateToProps `dispatch(fetchPrivilege(id))`

atcampbell: Just had a wee look and in other projects we declare initialise as a variable and just pass the parameter here:

`const initialise = ({ id }) => dispatch => {
    dispatch(fetchPrivilege(id));
};

const mapDispatchToProps {
  initialise
}`

src/Service.Host/client/src/containers/Privilege.js
```diff
@@ -0,0 +1,39 @@
...
+} from '../actions/privilegeActions';
+
+const getId = ownProps => ownProps.match.params.id;
+
+const mapStateToProps = (state, { match }) => ({
+    id: match.params.id,
+    privilege: getPrivilege(state)
+});
+
+const mapDispatchToProps = (dispatch, ownProps) => {
```
atcampbell: This can be simplified a bit

`const mapDispatchToProps = {
    initialise,
    togglePrivilegeStatus: togglePrivilegeStatus, etc...
}
`

src/Service.Host/client/src/middleware/itemCreated.js
```diff
@@ -0,0 +1,12 @@
+import { getSelfHref } from '@linn-it/linn-form-components-library';
+import history from '../history';
+
+export default () => next => action => {
+    const result = next(action);
+    if (action.type === 'RECEIVE_NEW_PRIVILEGE') {
+        history.push(getSelfHref(action.payload.data).slice(1));
```
atcampbell: What's the slice doing?

atcampbell:
I somehow managed to leave duplicate review comments...

src/Service.Host/client/src/middleware/itemCreated.js
iaincrawf: I had to take off a '/' from the front of the uri to link between some of the pages or it didn't generate the right urls, I think it generated the url without 'authorisation' in it maybe

iaincrawf:
> I somehow managed to leave duplicate review comments...

Haha, no worries! Do I just fix these and push again for you/someone to check?

src/Service.Host/client/src/containers/Privilege.js
lewisrenfrew: @atcampbell I thik this only works because of initialiseOnMount()  which isn't used here

src/Service.Host/client/src/containers/Privilege.js
iaincrawf: I have a feeling that this might also be because I haven't used initialiseOnMount, think it seems to need dispatch and ownprops for now until I swap it out later on

src/Service.Host/client/src/containers/Privilege.js
atcampbell: Ahhhh 👍👍👍👍

src/Service.Host/client/src/components/App.js
```diff
@@ -1,12 +1,34 @@
...
+    classes: PropTypes.shape({})
+};
+
+App.defaultProps = {
+    classes: {}
+};
 
-export default App;
\ No newline at end of file
+export default withStyles(styles)(App);
```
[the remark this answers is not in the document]
iaincrawf: going to carry this forward and try next week, need to upgrade to material ui 4 yeah and not sure what that involves

PR #15, opened 2019-06-27
update/save privilege implemented

iaincrawf:


src/Service.Host/client/src/components/Privilege.js
```diff
@@ -26,9 +26,9 @@ const ViewPrivilege = ({
         initialise();
     }, [initialise]);
 
-    // const handleSaveClick = () => {
-    //     savePrivilege(privilege.name, privilege.active);
-    // };
+    const handleSaveClick = () => {
+        savePrivilege(privilege.name, privilege.active, privilege.links[0].href);
```
richardiphillips: This is dangerous just accessing the link by index. Can fail if new links are added. 
You can use the getHref function in the shared library to get it by rel

PR #16, opened 2019-07-04
Privileges filtered by users + groups listed and able to be edited

iaincrawf:


src/Service.Host/client/src/actions/groupActions.js
```diff
@@ -0,0 +1,219 @@
...
+export const createGroup = (name, active) => ({
+    [RSAA]: {
+        endpoint: `${config.appRoot}/authorisation/groups`,
+        method: 'POST',
+        options: { requiresAuth: false },
+        headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json'
+        },
+        body: JSON.stringify({ name, active }),
```
lewisrenfrew: what's this line doin?

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -40,6 +49,57 @@ const ViewPrivilege = ({
     const privilegeName = privilege.name;
     const privilegeActive = privilege.active || false;
 
+    let elementsToDisplay;
+
```
lewisrenfrew: this elementsToDisplay variable seems kinda pointless.... why not just put all this in the return of the function?

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -21,43 +22,126 @@ const styles = () => ({
     }
 });
```
lewisrenfrew: I'd remove this styles thingy if its not in used

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -21,43 +22,126 @@ const styles = () => ({
...
+    selectedUser,
+    showCreate,
+    privilegesForAssignment,
+    createPermission,
+    currentUserUri
 }) => {
     useEffect(() => {
-        initialise();
+        initialise(selectedUser);
     }, [initialise]);
```
lewisrenfrew: should you put selectedUser in the array arg too??

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -128,13 +254,29 @@ ViewPrivileges.propTypes = {
...
+    currentUserUri: PropTypes.string.isRequired
 };
 
 ViewPrivileges.defaultProps = {
     classes: {},
     privileges: null,
-    newprivilege: { name: '', active: false }
+    newprivilege: { name: '', active: false },
+    users: [{ displayText: 'All users', id: -1 }, { displayText: 'test user', id: 12 }]
 };
```
lewisrenfrew: should this test data be coded in here as a default prop??

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -21,43 +22,126 @@ const styles = () => ({
...
+                        style={{
+                            height: '75px',
+                            maxWidth: '100px',
+                            borderRadius: '10px',
+                            overflow: 'hidden',
+                            display: 'inline-block',
+                            position: 'relative',
+                            left: '20px',
+                            top: '32px'
+                        }}
```
lewisrenfrew: seems kinda annoying to have all these styles inline. It's good to keep all of this stuff out of the return so it's easier to read/understand render logic.

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -83,27 +167,69 @@ const ViewPrivileges = ({
...
+                                                borderColor: 'rgba(0, 0, 0, 0.23)',
+                                                borderRadius: '5px',
+                                                cursor: 'pointer'
+                                            }}
+                                        >
+                                            <option value="-1">Select Privilege</option>
+                                            {showPrivilegesForAssignment ? (
+                                                privilegesForAssignment.map(p => (
+                                                    <option value={p.name}>{p.name}</option>
+                                                ))
```
lewisrenfrew: will react complain about mapped element keys here??

src/Service.Host/client/src/containers/Group.js
```diff
@@ -0,0 +1,51 @@
...
+        toggleGroupStatus: () => {
+            dispatch(toggleGroupStatus());
+        },
+        saveGroup: (name, active, uri) => {
+            dispatch(saveGroup(name, active, uri));
+        },
+        setUpdatedMessageVisible: visible => {
+            dispatch(setUpdatedMessageVisible(visible));
+        }
+    };
```
lewisrenfrew: this syntax seems kinda verbose... We usually just pass the action to be dispatched in here,,, see a file like this one:

https://github.com/linn/products-maint/blob/master/src/Service.Host/client/src/containers/salesArticles/SalesArticle.js

src/Service.Host/client/src/containers/Privileges.js
```diff
@@ -4,23 +4,64 @@ import Privileges from '../components/Privileges';
...
+            dispatch(fetchPrivileges());
+        }
+        dispatch(fetchUsers());
+    },
     createPrivilege: name => {
         dispatch(createPrivilege(name, false));
     },
     updateNewPrivilege: name => {
         dispatch(updateNewPrivilege(name));
+    },
```
lewisrenfrew: again, weird stuff going on in mapDispatch to props, specifically logic like what dispatch to map based on id and things... I'd move this logic out of here to keep things clean

tests/Integration/Service.Tests/GroupModuleSpecs/ContextBase.cs
```diff
@@ -1,5 +1,7 @@
 ﻿namespace Linn.Authorisation.Service.Tests.GroupModuleSpecs
 {
+    using System.Collections.Generic;
+
```
lewisrenfrew: evil blank line

src/Service.Host/client/src/components/Privileges.js
iaincrawf: yep good spot forgot to add them in!

src/Service.Host/client/src/containers/Group.js
iaincrawf: Think this is what we discussed the other day that I need to do it like this unless I use the initialise on mount magic

src/Service.Host/client/src/components/Privilege.js
iaincrawf: I actually kinda like this bit, splitting sorta logic on when things show away from the return and making it a bit easier to read (imo), buuut I haven't done it consistently at all and maybe since you guys normally don't I'm as well moving this one into the return too! Or go the other way and consistently make the rest like this

src/Service.Host/client/src/actions/groupActions.js
iaincrawf: I'm not 100% sure why but basically I get "SerializationException: Invalid JSON string" without it so added it in, it's just making it into Json but I'm likely being silly and just missing something! Will give you a shout at some point to see if you can see what I'm doing wrong or if I do need it

src/Service.Host/client/src/components/Privileges.js
iaincrawf: nope haha thanks

PR #18, opened 2019-07-30
authorisation - revoke privileges from group + delete member from group

iaincrawf:


src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
     };
 
+    const changeUser = e => {
+        selectUser(e.target.value);
+    };
+
+    const getEmployeeName = uri => {
+        if (members && users) {
+            const employee = users.find(x => x.href === uri);
+            if (employee) return employee.fullName;
```
atcampbell: could be `return employee ? employee.fullName : 'Employee not found'`. Maybe a bit clearer and you would need that second return statement below

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
-                                        }
-                                        label="Active Status"
-                                        labelPlacement="start"
-                                    />
-                                </Grid>
-                            </div>
-                        </Fragment>
-                    </Grid>
+    const editGroupSection = (
+        <div className={classes.bottomPadding}>
```
atcampbell: Is this div just being used for the padding? If so you could add that style to the Grid and remove the div

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+            <Grid container spacing={24}>
+                <Grid item xs={12}>
+                    <Fragment>
+                        <div>
```
atcampbell: You won't need this div as you're nesting the grid items in a fragment

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+        </div>
+    );
 
-    return (
+    const privilegesElements = (
+        <div className={classes.bottomPadding}>
+            <SubTitle>Privileges assigned to '{group.name}'</SubTitle>
+            {privileges && (
+                <div>
+                    <div className={classes.paddedScrollableDiv}>
```
atcampbell: Could these divs could be removed and the styles added to the table?

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+                </div>
+            )}
+        </div>
+    );
+
+    const createTable = membersList => {
+        const table = [];
+        for (let i = 0; i < membersList.length; i += 4) {
+            const children = [];
+            for (let j = 0; j < 4; j += 1) {
```
atcampbell: I think it would be clearer to use array methods here rather than nested for loops. 4 here seems like a magic number, what's its significance? I think the reliance on the i and j variables makes this a bit hard to follow exactly what's going on.

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+                                <Button
+                                    type="button"
+                                   // variant="outlined"
+                                    size="small"
+                                    onClick={() =>
+                                        dispatchRemoveMember(getHref(membersList[i + j], 'self'))
+                                    }
+                                    id="RemoveMember"
+                                >
+                                    X
```
atcampbell: Could the DeleteIcon go here?

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+                    >
+                        Assign to group
+                    </Button>
+                </div>
+            )}
+        </div>
+    );
+
+    const createTable = membersList => {
+        const table = [];
```
atcampbell: Could you construct a material ui table here instead of using html elements?

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+                    break;
+                }
+            }
+            table.push(<tr>{children}</tr>);
+        }
+        return table;
+    };
+
+    const membersElements = (
         <div>
```
atcampbell: Could be replaced with a fragment. Generally I think fragments are preferable for holding a bunch of react objects as divs add unneeded elements to the dom

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+            <SubTitle>Members of '{group.name}'</SubTitle>
+            {members && users && (
+                <div>
+                    <ol className={classes.paddedScrollableDiv}>
```
atcampbell: Could be a material list, or could be replaced with the MUI table?

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+                        <table style={{ width: '100%' }}>{createTable(members)}</table>
+                    </ol>
+
+                    <select
```
atcampbell: There's a dropdown component in the shared library that might work here

src/Service.Host/client/src/components/Group.js
```diff
@@ -46,83 +111,256 @@ const ViewGroup = ({
...
+                        value={selectedUser}
+                        defaultValue={-1}
+                        onChange={changeUser}
+                        className={classes.thinPrivilegesSelectList}
+                    >
+                        <option value={-1} key="-1">
```
atcampbell: I think these -1s shouldn't be hard coded, this option won't need a custom key as its not being created in a loop

src/Service.Host/client/src/components/Groups.js
```diff
@@ -12,18 +22,30 @@ const styles = () => ({
     }
 });
 
-const ViewGroups = ({ classes, initialise, groups, loading }) => {
+const ViewGroups = ({
```
atcampbell: These Components might be better as using the function declaration style to help debugging `function ViewGroups({ classes}) { return 'hiya' }`

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -21,19 +33,24 @@ const styles = () => ({
...
     useEffect(() => {
-        initialise();
-    }, [initialise]);
+        fetchPrivilege(id);
+      //  fetchEmployeesWithPrivilege(id),
+      fetchUsers();
+    }, [fetchPrivilege, fetchUsers, id]);
 
     const handleSaveClick = () => {
         savePrivilege(privilege.name, privilege.active, getHref(privilege, 'self'));
```
atcampbell: There's also a getSelfHref function so you don't need to pass in 'self'

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -77,9 +78,10 @@ const ViewPrivileges = ({
...
     currentUserUri,
-    //deletePermission,
+    deletePermission,
     showPrivilegeMessage,
-    setPrivilegeMessageVisible
+    setPrivilegeMessageVisible,
+    permissionMessage
 }) => {
     useEffect(() => {
         if (selectedUser == -1) {
```
atcampbell: Still a bit unsure about what -1 is being used for here, can you identify a user by their href?

src/Service/Modules/PrivilegesModule.cs
```diff
@@ -41,10 +42,17 @@ private object CreatePrivilege()
             return this.Negotiate.WithModel(result);
         }
 
+        private object RemovePrivilege(int id)
+        {
+            var resource = this.Bind<PrivilegeResource>();
+            //var result = this.privilegeService.RemovePrivilege(id, resource);
+            //return this.Negotiate.WithModel(result);
+            return null;
```
atcampbell: Something should be returned here. Maybe the user that has just been removed?

src/Domain/Services/PrivilegeService.cs
```diff
@@ -22,11 +25,13 @@ public IEnumerable<Privilege> GetPrivileges(string who)
         {
             // TODO what if who is string.Empty ?
```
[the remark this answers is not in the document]
iaincrawf: I didn't actually write this method but will add something in!

src/Facade/ResourceBuilders/PermissionResourceBuilder.cs
```diff
@@ -14,17 +14,19 @@ public PermissionResource Build(Permission permission)
                            {
                                GrantedByUri = individualPermission.GrantedByUri,
                                GranteeUri = individualPermission.GranteeUri,
-                               Privilege = individualPermission.Privilege.Name
-                           };
+                               Privilege = individualPermission.Privilege.Name,
+                               DateGranted = individualPermission.DateGranted.ToString("ddd, dd MMMM yyyy hh:mm tt")
```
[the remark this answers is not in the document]
iaincrawf: Cool will do

src/Service.Host/client/src/components/Group.js
```diff
@@ -1,8 +1,20 @@
...
+    Grid,
+    Switch,
+    Table,
+    TableHead,
+    TableBody,
+    TableCell,
+    TableRow
+} from '@material-ui/core';
+
+// import DeleteIcon from '@material-ui/icons/Delete';
```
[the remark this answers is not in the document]
iaincrawf: Sorry, part of this stuff was mid thought and I just pushed so I could move onto the production thing with lewis - was considering using the delete icon for remove further down this

src/Service.Host/client/src/components/Group.js
iaincrawf: I'd intended it to also return 'employee not found' if members or users hadn't been populated for whatever reason, so that's why I left it outside of that if statement like this, but maybe that should be a different 'error finding employee' string or something and do what you suggested inside?

src/Service.Host/client/src/components/Group.js
iaincrawf: Basically wanted to put 4 names in each row of a table, but found it a pain to write this with closing tags for the <tr> etc so it's a loop that takes 4 at a time and wraps each in a <td> and the 4 in a <tr>

src/Service.Host/client/src/components/Privileges.js
iaincrawf: in this class I used -1 to represent the "all users" option in the dropdown, and I'm using it to know if I want to get all the global privileges or just the privileges that one user actually has the permission for. I'll change it to "allUsers" or something more obvious than -1

src/Service/Modules/PrivilegesModule.cs
iaincrawf: I shouldn't have committed this as it hasn't been fully implemented yet sorry - can't remove privileges yet this was just a start towards it

src/Service.Host/client/src/components/Privilege.js
iaincrawf: Cheers 😄

src/Service.Host/client/src/components/Group.js
atcampbell: Ahh yeah sorry didn't realise that

src/Facade/ResourceBuilders/PermissionResourceBuilder.cs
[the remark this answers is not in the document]
iaincrawf: Do you print it out for users in iso format or do you convert it into something on the front end?

PR #20, opened 2019-11-01
Privileges update

iaincrawf:


src/Facade/IPermissionService.cs
```diff
@@ -1,19 +1,19 @@
...
+    using System.Collections.Generic;
 
     public interface IPermissionService : IFacadeService<Permission, int, PermissionResource, PermissionResource>
     {
         IResult<Permission> CreatePermission(PermissionResource permission);
+
         IResult<IEnumerable<Permission>> GetImmediatePermissionsForGroup(int groupId);
 
-       // IEnumerable<Privilege> GetAllPrivilegesForGroup(int groupId);
+        // IEnumerable<Privilege> GetAllPrivilegesForGroup(int groupId);
```
atcampbell: Can remove this comment?

src/Facade/PermissionService.cs
```diff
@@ -73,6 +73,20 @@ public IResult<IEnumerable<Permission>> GetImmediatePermissionsForGroup(int grou
         //    return privileges.Where(p => p.Active).Distinct();
         //}
```
atcampbell: Commented out stuff could be removed?

src/Service.Host/client/src/actions/userActions.js
```diff
@@ -0,0 +1,33 @@
+﻿import { RSAA } from 'redux-api-middleware';
+import config from '../config';
+import * as actionTypes from './actionTypes';
+
+const fetchUsers = () => ({
+    [RSAA]: {
+        //${config.appRoot} can be changed to https://app-sys.linn.co.uk to see names locally
```
atcampbell: Could remove this comment.

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -63,123 +63,106 @@ function ViewPrivilege({
...
-    //                         </TableHead>
-    //                         <TableBody>
-    //                             {employeesWithPrivilege.map(employee => (
-    //                                 <TableRow key={employee.name}>
-    //                                     {/* link to privilege */}
+    const createTable = (membersList, isGroup) => {
+        const table = [];
+        for (let i = 0; i < membersList.length; i += 4) {
+            const children = [];
+            for (let j = 0; j < 4; j += 1) {
```
atcampbell: Could this be split up - maybe creating the list of members elsewhere and then passing them in to a MUI table to display? I think that would make it a lot clearer what's going on. 

Might break the dependence on the for loops as well

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -63,123 +63,106 @@ function ViewPrivilege({
...
+                if (membersList[i + j]) {
+                    children.push(
+                        <td>
```
atcampbell: Could use MUI components

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -63,123 +63,106 @@ function ViewPrivilege({
...
-                                                checked={privilegeActive}
-                                                onChange={handleTogglePrivilegeStatus}
-                                                color="primary"
-                                            />
-                                        }
-                                        label="Active Status"
-                                        labelPlacement="start"
+            <Fragment>
+                <Grid item xs={12}>
+                    <Fragment>
```
atcampbell: Don't think this fragment is needed as its not the root of the return

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -63,123 +63,106 @@ function ViewPrivilege({
...
-    //                     onClick={dispatchcreatePermission}
-    //                     id="createNewPermission"
-    //                 >
-    //                     Assign to group
-    //                 </Button>
-    //             </div>
-    //         )}
-    //     </div>
-    // );
+    const individualPermissions = usersWithPermission
```
atcampbell: Do you need to do the conditional here? Can't remember what filtering when nothing found returns but it might already be an empty array (or null)

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -63,123 +63,106 @@ function ViewPrivilege({
...
-                    </Grid>
+                                }
+                                label="Active Status"
+                                labelPlacement="start"
+                            />
+                        </Grid>
+                        <Grid item xs={8}>
+                            {usersWithPermission && (
+                                <Fragment>
+                                    <h4>Groups With permission:</h4>
```
atcampbell: MUI components would be ggod

src/Service.Host/client/src/components/Privilege.js
```diff
@@ -63,123 +63,106 @@ function ViewPrivilege({
...
+                                        <table style={{ width: '100%' }}>
+                                            {createTable(individualPermissions, false)}
+                                        </table>
+                                    </ol>
+                                </Fragment>
+                            )}
+                        </Grid>
+                    </Fragment>
                 </Grid>
+                <br />
```
atcampbell: Padding or spacing would be more in line with the other UIs

src/Service.Host/client/src/reducers/users.js
```diff
@@ -0,0 +1,27 @@
+import * as actionTypes from '../actions/actionTypes';
+//define the initial state
+const initialState = { data: {}, loading: false };
+
+//define a reducer with an initialized state action
```
atcampbell: These comments could be removed

src/Service.Host/client/src/selectors/privilegeSelectors.js
```diff
@@ -56,10 +56,13 @@ export const getSaveEnabled = state => {
     return enableSave || false;
 };
 
-export const getAllUsers = state => {
-    const { privileges } = state;
-    const { users } = privileges;
-    return users;
+export const getPermissionsForPrivilege = state => {
+    const { privilege } = state;
```
atcampbell: Might be clearer with the ? operator state?.privilege?.permissions

src/Service.Host/client/src/components/Privilege.js
iaincrawf: I stole this from my existing code for speed, the list of members is already passed in but the loops were just a way of getting it to display four members per line. Seems kind of hacky now, I'm sure there's probably a MUI thing I can use as you say so will look into that!

src/Service.Host/client/src/components/Privilege.js
iaincrawf: good shout, looked it up and if it finds nothing it just returns an empty array so that's perfect for this with no conditional!

PR #57, opened 2021-11-09
Authorisation NuGet/NPM changes

RossStewart:
Upgrade the Authorisation project's NPM/NuGet packages to avoid any security issues.


lewisrenfrew (approved):

RossStewart:
There's just a slight project with the Domain test project thats breaking the build - will sort.

iaincrawf:
I thiiink it looks like we're still missing some npm package updates that dependabot wants, webpack-dev-server, lodash etc if you get time later!

RossStewart:
Ahhh ok, I'll sort that first

src/Service.Host/package.json
```diff
@@ -60,19 +60,23 @@
     "url-loader": "^0.6.2",
     "webpack": "^3.6.0",
     "webpack-dev-middleware": "^1.12.2",
-    "webpack-dev-server": "^2.11.3",
```
lewisrenfrew, line 63: did dependabot suggest this major version number update?.. Maybe double check the client still runs locally via npm start before merging

RossStewart: Yeah, it was suggested here : https://github.com/linn/authorisation/pull/21

WPDS definitely runs at 2.11.3 (with all these other changes) but is currently failing with the new major version release

lewisrenfrew: That's annoying. I assumed it would suggest a higher version within the same major release that included some kind of security patch. 

It's definitely worth doing, but foing from 2.x.x to 3.x.x might be a bit of a pain (seems likely webpack itself will need to be updated too). Is it spitting out any kind of sensible error message?

RossStewart: Mmmm this is a PR from 2019 so maybe that was the most secure version at the time? We could always see if we could go higher?

Yeah, I get the impression WP will also need updated. The error message on NPM START is : 

```
> authorisation@1.0.0 start C:\Development\auth\src\Service.Host
> node server.js

C:\Development\auth\src\Service.Host\node_modules\webpack-dev-server\lib\Server.js:115
    const { compile, invalid, done } = compiler.hooks;
            ^

TypeError: Cannot destructure property 'compile' of 'compiler.hooks' as it is undefined.
    at addHooks (C:\Development\auth\src\Service.Host\node_modules\webpack-dev-server\lib\Server.js:115:13)
    at new Server (C:\Development\auth\src\Service.Host\node_modules\webpack-dev-server\lib\Server.js:128:5)
    at Object.<anonymous> (C:\Development\auth\src\Service.Host\server.js:5:1)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)
    at internal/main/run_main_module.js:17:47
```

richardiphillips (approved):

PR #85, opened 2024-01-22
post hook idea

lewisrenfrew:
- this is my first attempt at doing some data posting with react / not redux
- i decided to make it reusable as a hook, not sure if thats the best idea or not.. but seemed kind of  sensible for now since i needed to call some other hooks (useState, useSelector)
- main point to note is that even thought this is a sort of redux-free POST, I'm still relying on the redux store to fish out the auth token for now
- also included a silly throwaway example of how to use it

### DIFF
--- a/src/Service.Host/client/src/components/Privileges.js
+++ b/src/Service.Host/client/src/components/Privileges.js
@@ -2,11 +2,13 @@ import React, { useEffect, useState } from 'react';
 import Typography from '@mui/material/Typography';
 import { Page, Loading } from '@linn-it/linn-form-components-library';
 import Grid from '@mui/material/Grid';
+import Button from '@mui/material/Button';
 import List from '@mui/material/List';
 import ListItem from '@mui/material/ListItem';
 import config from '../config';
 import history from '../history';
 import useInitialise from '../hooks/useInitialise';
+import usePost from '../hooks/usePost';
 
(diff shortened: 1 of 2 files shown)


richardiphillips (approved):
don't see any reason not to hook post. 
I know we'll end pulling out lots of shared code if we go down this route but even now it might be nice to still wrap the endpoint up in some sort of itemTypes type thing to avoid repetition and typos etc?

lewisrenfrew:
> don't see any reason not to hook post. I know we'll end pulling out lots of shared code if we go down this route but even now it might be nice to still wrap the endpoint up in some sort of itemTypes type thing to avoid repetition and typos etc?

Yeah,  good idea

lewisrenfrew:
> > don't see any reason not to hook post. I know we'll end pulling out lots of shared code if we go down this route but even now it might be nice to still wrap the endpoint up in some sort of itemTypes type thing to avoid repetition and typos etc?
> 
> Yeah, good idea

My other idea was to fold optional GET, PUT, POST., PATCH and DELETE functions all into the one useAPi(itemType) style hook which could just take the itemType along with some other params... keeping it more in line with how we are doing things currently but with a bit less ceremony - not too sure yet though. The purpose of this PR was really just to give oliver something to work with for his next project task

PR #86, opened 2024-01-22
Post New Priv

RossStewart:
Small test and endpoints for creating new privileges. Spoke Oliver through the test and architecture of the request and how it is created.

### DIFF
--- a/src/Facade/Services/PrivilegeFacadeService.cs
+++ b/src/Facade/Services/PrivilegeFacadeService.cs
@@ -11,7 +11,6 @@ namespace Linn.Authorisation.Facade.Services
 
     public class PrivilegeFacadeService : FacadeResourceService<Privilege, int, PrivilegeResource, PrivilegeResource>
     {
-
         public PrivilegeFacadeService(IRepository<Privilege, int> repository, ITransactionManager transactionManager, IBuilder<Privilege> resourceBuilder) 
             : base(repository, transactionManager, resourceBuilder)
         {
--- a/src/Facade/Services/PrivilegeFacadeService.cs
+++ b/src/Facade/Services/PrivilegeFacadeService.cs
@@ -19,11 +19,13 @@ namespace Linn.Authorisation.Facade.Services
         
         protected override Privilege CreateFromResource(PrivilegeResource resource, IEnumerable<string> privileges = null)
         {
-            return new Privilege
-            {
-                Id = resource.Id,
-                Name = resource.Name
-            };
+            var privilege = new Privilege 
+                                {
+                                    Id = resource.Id,
+                                    Name = resource.Name,
+                                    Active = resource.Active
+                                };
+            return privilege;
         }
         
         protected override void UpdateFromResource(Privilege entity, PrivilegeResource updateResource, IEnumerable<string> privileges = null)
--- a/src/Facade/Services/PrivilegeFacadeService.cs
+++ b/src/Facade/Services/PrivilegeFacadeService.cs
@@ -20,9 +20,7 @@ namespace Linn.Authorisation.Facade.Services
         {
             var privilege = new Privilege 
                                 {
-                                    Id = resource.Id,
-                                    Name = resource.Name,
-                                    Active = resource.Active
+                                    Name = resource.Name
                                 };
             return privilege;
         }
--- a/src/Facade/Services/PrivilegeFacadeService.cs
+++ b/src/Facade/Services/PrivilegeFacadeService.cs
@@ -19,7 +19,11 @@ namespace Linn.Authorisation.Facade.Services
         
         protected override Privilege CreateFromResource(PrivilegeResource resource, IEnumerable<string> privileges = null)
         {
-            throw new NotImplementedException();
+            return new Privilege
+            {
+                Id = resource.Id,
+                Name = resource.Name
+            };
         }
         
         protected override void UpdateFromResource(Privilege entity, PrivilegeResource updateResource, IEnumerable<string> privileges = null)
--- a/src/Facade/Services/PrivilegeFacadeService.cs
+++ b/src/Facade/Services/PrivilegeFacadeService.cs
@@ -18,10 +18,7 @@ namespace Linn.Authorisation.Facade.Services
         
         protected override Privilege CreateFromResource(PrivilegeResource resource, IEnumerable<string> privileges = null)
         {
-            var privilege = new Privilege 
-                                {
-                                    Name = resource.Name
-                                };
+            var privilege = new Privilege(resource.Name);
             return privilege;
         }
         
--- a/tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
+++ b/tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
@@ -5,9 +5,9 @@ namespace Linn.Authorisation.Integration.Tests.PrivilegeModuleTests
 
     using FluentAssertions;
 
+    using Linn.Authorisation.Domain;
     using Linn.Authorisation.Integration.Tests.Extensions;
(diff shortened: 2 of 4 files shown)


src/Facade/Services/PrivilegeFacadeService.cs
```diff
@@ -11,15 +11,20 @@ namespace Linn.Authorisation.Facade.Services
...
         }
         
         protected override Privilege CreateFromResource(PrivilegeResource resource, IEnumerable<string> privileges = null)
         {
-            throw new NotImplementedException();
+            var privilege = new Privilege 
+                                {
+                                    Id = resource.Id,
+                                    Name = resource.Name,
+                                    Active = resource.Active
```
lewisrenfrew, line 25: Actually a task for @OliverMclachlan in the morning, can we write a constructor for the Privilege domain class and pass in the name as a parameter instead of this object initializer syntax? you won't need to pass any Active since we just want that to default to true.. i.e. we can just set it as true in the new constructor you add, so all privileges are guaranteed to be born active.

src/Facade/Services/PrivilegeFacadeService.cs
```diff
@@ -11,15 +11,20 @@ namespace Linn.Authorisation.Facade.Services
...
             : base(repository, transactionManager, resourceBuilder)
         {
         }
         
         protected override Privilege CreateFromResource(PrivilegeResource resource, IEnumerable<string> privileges = null)
         {
-            throw new NotImplementedException();
+            var privilege = new Privilege 
+                                {
+                                    Id = resource.Id,
```
lewisrenfrew, line 23: We won't know the Id yet - the database will generate one for us so can kill this line

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
```diff
@@ -0,0 +1,49 @@
...
+    using NUnit.Framework;
+
+    public class WhenPosting : ContextBase
+    {
+        private PrivilegeResource resource;
+
+        [SetUp]
+        public void SetUp()
+        {
+            this.resource = new PrivilegeResource { Id = 1, Name = "1", Active = true };
```
lewisrenfrew, line 23: doesnt matter massively, but wouldn't pass and Id or an Active flag here since we don't know Id yet / active will just default to true once we add our new constructor

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
```diff
@@ -0,0 +1,49 @@
...
+            this.Response.StatusCode.Should().Be(HttpStatusCode.Created);
+        }
+
+        [Test]
+        public void ShouldReturnJsonContentType()
+        {
+            this.Response.Content.Headers.ContentType.Should().NotBeNull();
+            this.Response.Content.Headers.ContentType?.ToString().Should().Be("application/json");
+        }
+
```
lewisrenfrew, line 40: I would also add a test to check the repository.Add() is being called with the correct data (i.e. just a new privilege with a name field matching the one on the resource)

src/Facade/Services/PrivilegeFacadeService.cs
lewisrenfrew: worth double checking this happens in POSTman or something if you haven't already

lewisrenfrew (changes requested):

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
RossStewart: Made the suggested changes in latest commit. Added a ShouldCallAdd test too but since PrivilegeRepository is currently subbed out I've looked at ANY and changed checking the ID at the bottom to now be the name. If we choose to not sub it out, I can always change this but gives Oliver a good place to get going

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
```diff
@@ -20,7 +20,7 @@ public class WhenPosting : ContextBase
         [SetUp]
         public void SetUp()
         {
-            this.resource = new PrivilegeResource { Id = 1, Name = "1", Active = true };
+            this.resource = new PrivilegeResource { Name = "1" };
```
richardiphillips, line 23: this test would be more readable if the name doesn't look like a number

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
richardiphillips: Choosing not to stub out a repository would be a wild suggestion. You can test that the add is called with the correct object though i.e. has the correct name.

richardiphillips: you could also mock the add action to update the object with an id but there's no need really as you can assume that the common repo is tested elsewhere.

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void ShouldReturnJsonContentType()
+        {
+            this.Response.Content.Headers.ContentType.Should().NotBeNull();
+            this.Response.Content.Headers.ContentType?.ToString().Should().Be("application/json");
+        }
+
+        [Test]
+        public void ShouldReturnJsonBody()
+        {
+            var resources = this.Response.DeserializeBody<PrivilegeResource>();
```
richardiphillips, line 50: resource rather than resources would indicate that you're only expecting one thing back

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
```diff
@@ -0,0 +1,55 @@
...
+            resources.Should().NotBeNull();
```
richardiphillips, line 51: nothing wrong with this but my personal opinion would be that it's not needed.
it's tested by the next line and there's no need to be defensive in tests.

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
richardiphillips: Deleted my previous comment as it didn't add anything of interest to Oliver

tests/Integration/Integration.Tests/PrivilegeModuleTests/WhenPosting.cs
RossStewart: No worries. Will remove. I quite like these in a testing set up as it requires less inspecting of the result when it comes back but can understand why it's redundant.

RossStewart: Updates made. If someone okays them I'll make sure to merge them in before start of tomorrow.

src/Facade/Services/PrivilegeFacadeService.cs
```diff
@@ -11,15 +11,18 @@ namespace Linn.Authorisation.Facade.Services
...
-
         public PrivilegeFacadeService(IRepository<Privilege, int> repository, ITransactionManager transactionManager, IBuilder<Privilege> resourceBuilder) 
             : base(repository, transactionManager, resourceBuilder)
         {
         }
         
         protected override Privilege CreateFromResource(PrivilegeResource resource, IEnumerable<string> privileges = null)
         {
-            throw new NotImplementedException();
+            var privilege = new Privilege
```
lewisrenfrew, line 21: We already have the constructor we need, so just use it here

PR #87, opened 2024-01-29
Create privilege story

OliverMclachlan:
Not a proper PR - Just looking at the differences in code


src/Service.Host/client/src/components/Privileges.js
```diff
@@ -37,7 +33,7 @@ function Privileges() {
     const renderPrivilege = privilege => {
         if (privilege.active === true) {
             return (
-                <ListItem key={privilege.id}>
+                <ListItem component={Link} to="/authorisation/Id" key={privilege.id}>
```
lewisrenfrew, line 36: I think changing this component to be a "Link" changes its role in the document, and so the test query
       
 const listItems = getAllByRole('listitem');

no longer returns the ACTIVE items which will now have a different role....

 Another example of react tests being quite picky about exact choice of elements, text etc. which I know is a bit of a pain. 

We can look into better ideas in the future, but I think for now you should just revert any changes to this file and try to keep this branch solely concerned with your "create privilege" story.

We don't want situations like this to arise where tests relating to a completely different story are failing and blocking our progess on this one, so one story per branch is a generally good idea if we're being strict about our test-driven agile approach

OliverMclachlan: Ok. Thanks for the help

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,122 @@
...
+import Snackbar from '@mui/material/Snackbar';
+
+import useInitialise from '../hooks/useInitialise';
+import usePost from '../hooks/usePost';
+import config from '../config';
+import history from '../history';
+
+function CreatePrivilege() {
+    const [inputValue, setInputValue] = useState('');
+    const [privileges, setPrivileges] = useState([]);
```
lewisrenfrew, line 15: I don't think we need a state item for privileges here since its just a readonly list of privileges. can just use the 'data' value directly as opposed to loading it into state?

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,122 @@
...
+
+    const spinningWheel = () => {
+        if (isLoading) {
+            return <Loading />;
+        }
+        return <div />;
+    };
+
+    // any time data changes, load it into our local privileges state
+    useEffect(() => {
```
lewisrenfrew, line 36: likewise we won't need this useEffect, just use the value of data directly on line 44 for your forEach loop check, e.g. data?.forEach(...)

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,122 @@
...
+                        createNewPrivilege(inputValue.trim());
+                    }}
+                    disabled
+                >
+                    Save
+                </Button>
+            );
+        }
+        return (
+            <Button
```
lewisrenfrew, line 77: we don't need two seperate returns and two seperate buttons here. We can just have one buttong and set the disabled prop to be true or false depending on the presnece of an 'input' value

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,122 @@
...
+    };
+
+    // any time data changes, load it into our local privileges state
+    useEffect(() => {
+        if (data) {
+            setPrivileges(data);
+        }
+    }, [data]);
+
+    function isNameValid(name) {
```
lewisrenfrew, line 42: this is above and beyond the spec! which is a good thing, but means there will be no tests to cover this logic.... We could always write some....

from a glance I can spot one small issue in that the check returns true/valid if no privileges have been returned from the API yet, so it is possible that the user could sneak in and create a duplicately-named privilege quickly, before the existing ones are loaded, e.g. if the server was being slow to respond with the existing privileges for some reason. 

don't worry about it too much though - we can talk about a more robust method for input validation later on

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,122 @@
...
+        let valid = true;
+        privileges.forEach(
+            privilege => {
+                if (name === privilege.name) {
+                    valid = false;
+                }
+            },
+            [valid]
```
lewisrenfrew, line 50: whats this second array param being passed to forEach() doing? not seen that before

lewisrenfrew (changes requested):
- looking good - few small suggestions of things to change before we move on

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,84 @@
...
+    );
+
+    const spinningWheel = () => {
+        if (isLoading) {
+            return <Loading />;
+        }
+        return <div />;
+    };
+
+    // any time data changes, load it into our local privileges state
```
lewisrenfrew, line 35: can delete this comment

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,84 @@
...
+
+    function createNewPrivilege() {
```
lewisrenfrew, line 37: I think we can delete this function now, and just call pass send directly to our buttons onClick

src/Service.Host/client/src/components/CreatePrivilege.js
```diff
@@ -0,0 +1,84 @@
...
+                        propertyName="inputValue"
+                        label="Name"
+                        value={inputValue}
+                        onChange={handleFieldChange}
+                        fullWidth
+                    />
+
+                    <Button
+                        variant="contained"
+                        onClick={() => {
```
lewisrenfrew, line 61: e.g. onClick={send}

lewisrenfrew (approved):

Issue #83, opened 2024-01-18
User should be able to click on a privilege from the main list and be directed to a screen showing data specific to that Privilege


lewisrenfrew:
- The screen should consist of a new Privilege component, rendered at authorisation/privileges/id , where id is the numeric id of the privilege that the user clicked
- the component should fetch the individual privilege data from the api when it first renders
- The component Should just render the name and active status of the privilege in question
- should render a loading spinner when data is loading
- The user should be able to navigate to the new page by selecting and clicking a privilege from the main Privileges compononet, which should then link them to the specific privilege they clicked
- both ACTIVE and INACTIVE priviliges should be clickable links

lewisrenfrew:
back end work still outstanding

PR #88, opened 2024-01-31
View privilege story

OliverMclachlan:
Individual privileges page #83


src/Service.Host/client/src/components/Privilege.js
```diff
@@ -0,0 +1,53 @@
...
+    };
+
+    return (
+        <Page homeUrl={config.appRoot} history={history}>
+            <Grid container spacing={3}>
+                <Grid item xs={12}>
+                    {spinningWheel()}
+                </Grid>
+                <Grid item xs={12}>
+                    <List>
```
lewisrenfrew, line 40: I wouldn't bother with a List here since there is only one. Just the Typography component on its own will suffice

lewisrenfrew (approved):

PR #90, opened 2024-02-02
Go nuclear with redux hopefully

lewisrenfrew:
- rip out redux
- replace redux-oidc implict flow auth with react-context-oidc authorization code flow auth, which is much more modern and secure
- need a new auth client to support the new grant type, so I've just created app2 as a client in the sys db for now
- really just a proof of concept PR for demonstration purposes, can talk through it all at next show and tell


richardiphillips (approved):
looks good. 
improving the auth could end up being the death of redux. 
I think the item created middleware is still useful though?

lewisrenfrew:
> looks good. improving the auth could end up being the death of redux. I think the item created middleware is still useful though?

as in the ability to just do something after a succesful POST? could always add an onSuccess() routine to the usePost hook or something

richardiphillips:
Yeah good idea. It's just such a common thing that you'll want to navigate to display the thing that's just been created.

PR #94, opened 2024-02-15
View permissions

OliverMclachlan:
Ability to view an employee's permissions


src/Service.Host/client/src/components/ViewPermission.js
```diff
@@ -0,0 +1,73 @@
...
+                        setEmployeeInput(newValue);
+                        send(null, `?who=/employees/${newValue}`);
+                    }}
+                    value={employeeInput}
+                />
+            </Grid>
+
+            <Grid item xs={12}>
+                <List>{permissions?.map(renderEmployeesPermission)}</List>
+            </Grid>
```
lewisrenfrew, line 68: react will probably be complaining about list items not having a key prop, you should probably pass the privilege.id or something to the renderEmployeesPermission function so that you can give each ListItem a unique key={privilegeId} prop

lewisrenfrew (approved):
nice

src/Service.Host/Program.cs
```diff
@@ -13,7 +13,7 @@ public static void Main(string[] args)
         public static IWebHost BuildWebHost(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
                 .UseStartup<Startup>()
-                .UseUrls("http://+:5050")
+                .UseUrls("http://+:51799")
```
lewisrenfrew, line 16: don't forget to revert to 5050 before you merge

RossStewart:
> I think VisualStudioCode is adding this file for some reason - same happening in sales. Not sure why this is happening but can you add it to the gitignore before merging this PR? That'll stop it sneaking into PRs at least

Was just talking to Oliver about the .gitignore stuff. It's a really handy tool to have for situations exactly like this. If you want me to show you it just call Oliver!

PR #96, opened 2024-02-21
Test passing

OliverMclachlan:
Duplicate permissions should be stopped

### DIFF
--- a/src/Domain/Permissions/IndividualPermission.cs
+++ b/src/Domain/Permissions/IndividualPermission.cs
@@ -1,9 +1,8 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
-
 namespace Linn.Authorisation.Domain.Permissions
 {
+    using System.Collections.Generic;
+    using System.Linq;
+
     public class IndividualPermission : Permission
     {
         public IndividualPermission(string granteeUri, Privilege privilege, string grantedByUri) : base(privilege, grantedByUri)
@@ -19,13 +18,15 @@ namespace Linn.Authorisation.Domain.Permissions
         public string GranteeUri { get; set; }
 
         public bool CheckUnique(IEnumerable<IndividualPermission> existingPermissions)
-        {
-            foreach (var permission in existingPermissions.Where(p => p.Privilege.Id == this.Privilege.Id && p.GranteeUri == this.GranteeUri))
+        { 
+            bool isUnique = true;
+      
+            foreach (var permission in existingPermissions.Where(p => p.Privilege.Name == this.Privilege.Name && p.GranteeUri == this.GranteeUri))
             {
-                return false;
+                isUnique = false;
             }
 
-            return true;
+            return isUnique;
         }
     }
 }
--- a/src/Domain/Permissions/IndividualPermission.cs
+++ b/src/Domain/Permissions/IndividualPermission.cs
@@ -1,3 +1,7 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
 namespace Linn.Authorisation.Domain.Permissions
 {
     public class IndividualPermission : Permission
@@ -12,6 +16,16 @@ namespace Linn.Authorisation.Domain.Permissions
             // empty args constructor needed for ef
         }
 
-        public string GranteeUri { get; set; }       
+        public string GranteeUri { get; set; }
+
+        public bool CheckUnique(IEnumerable<IndividualPermission> existingPermissions)
+        {
+            foreach (var permission in existingPermissions.Where(p => p.Privilege.Id == this.Privilege.Id && p.GranteeUri == this.GranteeUri))
+            {
+                return false;
+            }
+
+            return true;
+        }
     }
 }
--- a/src/Domain/Permissions/IndividualPermission.cs
+++ b/src/Domain/Permissions/IndividualPermission.cs
@@ -1,8 +1,6 @@
 namespace Linn.Authorisation.Domain.Permissions
 {
     using System.Collections.Generic;
-    using System.Linq;
-    using System.Security;
 
     public class IndividualPermission : Permission
     {
--- a/src/Domain/Permissions/IndividualPermission.cs
+++ b/src/Domain/Permissions/IndividualPermission.cs
@@ -21,9 +21,12 @@ namespace Linn.Authorisation.Domain.Permissions
         { 
             bool isUnique = true;
       
-            foreach (var permission in existingPermissions.Where(p => p.Privilege.Name == this.Privilege.Name && p.GranteeUri == this.GranteeUri))
+            foreach (var permission in existingPermissions)
             {
-                isUnique = false;
+                if (permission.Privilege.Name == this.Privilege.Name && permission.GranteeUri == this.GranteeUri)
+                { 
+                    isUnique = false; 
+                }
             }
 
             return isUnique;
--- a/src/Domain/Permissions/IndividualPermission.cs
+++ b/src/Domain/Permissions/IndividualPermission.cs
@@ -2,6 +2,7 @@ namespace Linn.Authorisation.Domain.Permissions
 {
     using System.Collections.Generic;
     using System.Linq;
+    using System.Security;
 
     public class IndividualPermission : Permission
     {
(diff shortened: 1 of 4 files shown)


src/Domain/Permissions/IndividualPermission.cs
```diff
@@ -12,6 +16,16 @@ public IndividualPermission()
             // empty args constructor needed for ef
         }
 
-        public string GranteeUri { get; set; }       
+        public string GranteeUri { get; set; }
+
+        public bool CheckUnique(IEnumerable<IndividualPermission> existingPermissions)
+        {
+            foreach (var permission in existingPermissions.Where(p => p.Privilege.Id == this.Privilege.Id && p.GranteeUri == this.GranteeUri))
```
lewisrenfrew, line 23: seems like quite a strange and convoluted way to do this check.

specifically, it seems very odd to have a foreach loop that just returns false on its first ever iteration - a loop that never loops shouldnt be a loop.. if that makese sense!

src/Domain/Permissions/Permission.cs
```diff
@@ -1,6 +1,10 @@
 namespace Linn.Authorisation.Domain.Permissions
 {
     using System;
+    using System.Collections.Generic;
+    using System.ComponentModel.DataAnnotations;
```
lewisrenfrew, line 5: delete all these

src/Domain/Permissions/Permission.cs
```diff
@@ -21,5 +25,9 @@ protected Permission(Privilege privilege, string grantedByUri)
         public DateTime DateGranted { get; set; }
 
         public string GrantedByUri { get; set; }
+
+        
+
```
lewisrenfrew, line 30: delete blank lines

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
...
+    public class WhenCheckingUniqueAndNotUnique : ContextBase
+    {
+        private readonly string privilegeName = "do.admin.stuuuff";
+        private List<IndividualPermission> permissions;
+        private IndividualPermission individualPermission;
+
+        [SetUp]
+        public void SetUp()
+        {
+            individualPermission = new IndividualPermission
```
lewisrenfrew, line 20: should use this.individaulPermission anywhere you refer to individualPermission field (since its a class field of this test class) for clarity

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                GranteeUri = "/employees/133",
+                Privilege = new Privilege(this.privilegeName),
+                GrantedByUri = "/employees/7004",
+                DateGranted = DateTime.UtcNow
+            };
+
+            permissions = new List<IndividualPermission>
```
lewisrenfrew, line 28: same here, should be this.permissions = 

again just for clairty

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                new IndividualPermission{
+                GranteeUri = "/employees/133",
+                Privilege = new Privilege(this.privilegeName),
+                GrantedByUri = "/employees/7004",
+                DateGranted = DateTime.UtcNow
+                }
+                ,
```
lewisrenfrew, line 36: weird formatting

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
...
+        private List<IndividualPermission> permissions;
+        private IndividualPermission individualPermission;
+
+        [SetUp]
+        public void SetUp()
+        {
+            individualPermission = new IndividualPermission
+            {
+                GranteeUri = "/employees/133",
+                Privilege = new Privilege(this.privilegeName),
```
lewisrenfrew, line 23: Privilege Id field needs to be assigned- it is the Id field that your code is using to check whether they are already in use in an existingPermission for this Grantee. I think this test only passes as it stands because Privilege Id happens to be null for both the IndividualPermission you are testing, and every Permission in your existing permissions list.. which makes this kind of a dodgy test

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
...
+                new IndividualPermission{
+                GranteeUri = "/employees/133",
+                Privilege = new Privilege(this.privilegeName),
+                GrantedByUri = "/employees/7004",
+                DateGranted = DateTime.UtcNow
+                }
+                ,
+                new IndividualPermission{
+                GranteeUri = "/employees/3006",
+                Privilege = new Privilege(this.privilegeName),
```
lewisrenfrew, line 39: same here, we need Privilege to have its Id field set for this test to make sense

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
...
+            };
+            this.PermissionRepository.FilterBy(Arg.Any<Expression<Func<Permission, bool>>>())
+                .Returns(permissions.AsQueryable());
+        }
+            [Test]
+            public void ShouldReturnFalse()
+            {
+                var result = individualPermission.CheckUnique(permissions);
+
+                Assert.IsFalse(result);
```
lewisrenfrew, line 52: Conventionally we use the fluent assertions syntax,
so something like result.should.beFalse()

lewisrenfrew (commented):
I think adding another test to check the happy returns true scenario would be good practise (once you fix the existing one)

src/Domain/Permissions/IndividualPermission.cs
```diff
@@ -1,3 +1,7 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
lewisrenfrew, line 3: just spotted this too - usings should be inside the namespace

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Linn.Authorisation.Domain.Permissions;
+using NSubstitute;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
```
lewisrenfrew, line 7: usings should be inside the namespace

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
lewisrenfrew: on second thought, I'd be tempted to use the privilege name in you check anyway, since thats the important thing

src/Domain/Permissions/IndividualPermission.cs
lewisrenfrew: yeah, change your check to compare Privilege Names as opposed to Ids. We don't really care about the Id its the Name string that actually results in a user being permitted to do something

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
lewisrenfrew: can ignore these comments about setting Id if we change the code to check on Name instead

src/Domain/Permissions/IndividualPermission.cs
```diff
@@ -2,6 +2,7 @@ namespace Linn.Authorisation.Domain.Permissions
 {
     using System.Collections.Generic;
     using System.Linq;
+    using System.Security;
```
lewisrenfrew, line 5: think this using statement might have sneaked in

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUnique.cs
```diff
@@ -63,7 +63,7 @@ public void ShouldReturnFalse()
 
                 result.Should().BeFalse();
             }
```
lewisrenfrew, line 65: put  a blank line in here between these two test methods

lewisrenfrew (approved):

PR #97, opened 2024-02-22
Fixing map crash

OliverMclachlan:
Adding ? to prevent error from happening

### DIFF
--- a/src/Service.Host/client/src/components/CreatePermission.js
+++ b/src/Service.Host/client/src/components/CreatePermission.js
@@ -71,7 +71,7 @@ function CreatePermission() {
             <Grid item xs={4}>
                 <Dropdown
                     propertyName="privilege choice"
-                    items={privileges.map(privilege => ({
+                    items={privileges?.map(privilege => ({
                         id: privilege.id,
                         displayText: privilege?.name
                     }))}
@@ -85,7 +85,7 @@ function CreatePermission() {
             <Grid item xs={4}>
                 <Dropdown
                     propertyName="employee choice"
-                    items={employees.items.map(employee => ({
+                    items={employees?.items.map(employee => ({
                         id: employee.id,
                         displayText: `${employee?.firstName} ${employee?.lastName}`
                     }))}


src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -19,48 +20,95 @@ public class PermissionFacadeService : IPermissionFacadeService
...
             var permission = new IndividualPermission(
                 permissionResource.GranteeUri, privilege, permissionResource.GrantedByUri);
 
-            this.permissionsRepository.Add(permission);
+            var permissions = this.permissionsRepository.FilterBy(p => p is IndividualPermission);
+
+            var individualPermissions = permissions.Select(p => (IndividualPermission)p);
+
+            
+
```
lewisrenfrew, line 55: some blank lines here to be deleted

src/Service.Host/client/src/hooks/usePost.js
```diff
@@ -43,6 +43,9 @@ function usePost(url, id, data, requiresAuth = false) {
                 setIsLoading(false);
             });
     };
+
+
```
lewisrenfrew, line 47: some blanks

tests/Integration/Integration.Tests/PermissionsModuleTests/WhenPosting.cs
```diff
@@ -31,7 +31,7 @@ public void SetUp()
                                     GrantedByUri = "/employees/33156",
                                     GranteeUri = "/employees/33190",
                                     DateGranted = DateTime.Now.ToString("o"),
-                                    GranteeGroupId = null,
+                                    GranteeGroupId = 0,
```
lewisrenfrew, line 34: GranteeGroupId should be null here (if you had to change it to zero because we changed its type to int, just change its type to be a nullable int)

lewisrenfrew (approved):
looks good - I personally prefer one 'scenario' per test class as opposed to the two scenarios you've got in each test class - but not a big deal

tests/Unit/Domain.Tests/PermissionServiceTests/WhenCheckingUniqueAndNotUniqueGroup.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace Linn.Authorisation.Domain.Tests.PermissionServiceTests
```
lewisrenfrew, line 1: also, these two new test classes are currently in the 'PermissionServiceTests' folder and namespace. It would be nice to move them into their own PermissionTests folder and namespace

PR #100, opened 2024-02-28
Group permission posting

OliverMclachlan:
Adding a create group permission page


RossStewart:
@OliverMclachlan You want to revert the sln file here or else it won't build properly

lewisrenfrew:
> @OliverMclachlan You want to revert the sln file here or else it won't build properly



> @OliverMclachlan You want to revert the sln file here or else it won't build properly

did you not add this to the .gitignore? Still have no idea why this file keeps popping up. I think vscode is generating it but I have no idea why

PR #103, opened 2024-03-01
Group create group

OliverMclachlan:
Adding the ability to create a new group


src/Facade/Services/GroupFacadeService.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            this.groupRepository = groupRepository;
+        }
+
+        protected override Group CreateFromResource(GroupResource resource, IEnumerable<string> privileges = null)
+        {
+            var active = true;
+            var group = new Group(resource.Name, resource.Active);
+
+            var groups = this.groupRepository.FilterBy(g => g is Group);
```
lewisrenfrew, line 37: can just do groupRepository.FindAll()?

src/Facade/Services/GroupFacadeService.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (group.CheckUnique(groups))
+            {
+                this.groupRepository.Add(group);
+                this.transactionManager.Commit();
+
+                var result = new Group { Name = resource.Name, Active = resource.Active, };
```
lewisrenfrew, line 44: extra comma after active

src/Service.Host/client/src/components/CreateGroup.js
```diff
@@ -0,0 +1,72 @@
...
+                        propertyName="inputValue"
+                        label="Name"
+                        value={inputValue}
+                        onChange={handleFieldChange}
+                        fullWidth
+                    />
+
+                    <Button
+                        variant="contained"
+                        onClick={() => {
```
lewisrenfrew, line 49: again, just pass send reference directy to onClick for conciseness, onClick={send}

src/Facade/Services/GroupFacadeService.cs
lewisrenfrew: can also just return group here? dont need to new up an additional grou

lewisrenfrew (approved):

PR #104, opened 2024-03-14
Group edit

OliverMclachlan:
Adding the ability to edit groups and not allowing duplicate name

### DIFF
--- a/src/Domain/Groups/Group.cs
+++ b/src/Domain/Groups/Group.cs
@@ -13,6 +13,14 @@ namespace Linn.Authorisation.Domain.Groups
             this.Members = new List<Member>();
         }
 
+        public Group(string name, bool active, int id)
+        {
+            this.Name = name;
+            this.Active = active;
+            this.Id = id;
+            this.Members = new List<Member>();
+        }
+
         public Group()
         {
             this.Members = new List<Member>();
--- a/src/Domain/Groups/Group.cs
+++ b/src/Domain/Groups/Group.cs
@@ -67,6 +67,18 @@ namespace Linn.Authorisation.Domain.Groups
             return true;
         }
 
+        public bool CheckUpdatedNameIsUnique(IEnumerable<Group> existingGroups)
+        {
+            foreach (var group in existingGroups)
+            {
+                if (group.Name == this.Name && group.Id == this.Id)
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
         public void RemoveMember(Member member)
         {
             this.Members.Remove(member);
@@ -80,6 +92,12 @@ namespace Linn.Authorisation.Domain.Groups
         public bool IsMemberOf(string who)
         {
             return this.MemberUris().Contains(who);
+        }
+
+        public void Update(string name, bool active)
+        {
+            this.Name = name;
+            this.Active = active;
         }
     }
 }
--- a/src/Domain/Groups/Group.cs
+++ b/src/Domain/Groups/Group.cs
@@ -67,18 +67,6 @@ namespace Linn.Authorisation.Domain.Groups
             return true;
         }
 
-        public bool CheckUpdatedNameIsUnique(IEnumerable<Group> existingGroups)
-        {
-            foreach (var group in existingGroups)
-            {
-                if (group.Name == this.Name)
-                {
-                    return false;
-                }
-            }
-            return true;
-        }
-
         public void RemoveMember(Member member)
         {
             this.Members.Remove(member);
--- a/src/Domain/Groups/Group.cs
+++ b/src/Domain/Groups/Group.cs
@@ -71,7 +71,7 @@ namespace Linn.Authorisation.Domain.Groups
         {
             foreach (var group in existingGroups)
             {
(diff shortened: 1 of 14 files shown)


src/Domain/Groups/Group.cs
```diff
@@ -67,6 +67,18 @@ public bool CheckUnique(IEnumerable<Group> existingGroups)
             return true;
         }
 
+        public bool CheckUpdatedNameIsUnique(IEnumerable<Group> existingGroups)
+        {
+            foreach (var group in existingGroups)
+            {
+                if (group.Name == this.Name && group.Id == this.Id)
```
lewisrenfrew, line 74: probably just want to check the name here, not the Id too

src/Service/Modules/GroupModule.cs
```diff
@@ -23,6 +23,7 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
             endpoints.MapGet("/authorisation/groups", this.GetAll);
             endpoints.MapGet("/authorisation/groups/{id:int}", this.GetGroup);
             endpoints.MapPost("/authorisation/groups", this.CreateGroup);
+            endpoints.MapPut("/authorisation/groups/{id:int}", this.UpdateGroup);
```
lewisrenfrew, line 26: see if you can work out how to write an integration test for this new endpoint? will be pretty similiar to the integration tests for PUTting permissions and privileges, so check them for reference

lewisrenfrew: looks good - as an added bonus task you could try and work out how to write another test files (called something like 'WhenUpdatingNameAndNameAlreadyTaken.cs') that checks that a bad request result is returned if you try and PUT a duplicate group name? you can use this test for inspiration:
https://github.com/linn/authorisation/blob/master/tests/Integration/Integration.Tests/PermissionsModuleTests/WhenPostingAndPermissionAlreadyExistsForThisPrivilegeAndGrantee.cs

you'll just need to  add a repository.FilterBy().Returns mock in your test setup that contains a group that is already taking the name of the new name you are passing in your test resource

src/Facade/Services/GroupFacadeService.cs
```diff
@@ -11,6 +11,8 @@
     using Linn.Common.Facade;
     using Linn.Common.Persistence;
 
+    using Microsoft.EntityFrameworkCore.ChangeTracking;
```
lewisrenfrew, line 14: probably don't need this?

richardiphillips:
Nothing to do with this PR but just as general feedback - could serve the bundle up on a few more routes to allow the user to go straight to them without having to start at the front page. 
(similar to https://github.com/linn/authorisation/blob/5890a3e8da6f4c75cdd59a49662c8ecd6a558de7/src/Service/Modules/ApplicationModule.cs#L18)

e.g. if you try and hit https://app-sys.linn.co.uk/authorisation/permission/view/individual the backend has no route to serve it

richardiphillips:
also is there a reason that the client is using /group but the backend seems to favour /groups?

lewisrenfrew:
> Nothing to do with this PR but just as general feedback - could serve the bundle up on a few more routes to allow the user to go straight to them without having to start at the front page. (similar to
> 
> https://github.com/linn/authorisation/blob/5890a3e8da6f4c75cdd59a49662c8ecd6a558de7/src/Service/Modules/ApplicationModule.cs#L18
> 
> )
> e.g. if you try and hit https://app-sys.linn.co.uk/authorisation/permission/view/individual the backend has no route to serve it

because of the way the authentication works (or more, doesn't work!) you just get redirected back to the home page anyway as it stands. I am going to try fix that today and then I'll make sure the app is also served up at all the accessible routes

richardiphillips:
> > Nothing to do with this PR but just as general feedback - could serve the bundle up on a few more routes to allow the user to go straight to them without having to start at the front page. (similar to
> > https://github.com/linn/authorisation/blob/5890a3e8da6f4c75cdd59a49662c8ecd6a558de7/src/Service/Modules/ApplicationModule.cs#L18
> > 
> > )
> > e.g. if you try and hit https://app-sys.linn.co.uk/authorisation/permission/view/individual the backend has no route to serve it
> 
> because of the way the authentication works (or more, doesn't work!) you just get redirected back to the home page anyway as it stands. I am going to try fix that today and then I'll make sure the app is also served up at all the accessible routes

ahhh  forgot about the auth thing here

tests/Integration/Integration.Tests/Integration.Tests.csproj
RossStewart, line 1: I don't think this file needs to be edited?

RossStewart (approved):

tests/Integration/Integration.Tests/Integration.Tests.csproj
OliverMclachlan: Was just because i created 2 or 3 folder that i deleted later on and it just had code saying newFolder1 & newFolder2

RossStewart: Ah right no bother then!

PR #105, opened 2024-03-15
Api request todos

lewisrenfrew:
- redirect back to the url the user tried to sign in from in the onSignInCallback function . This took me ages to figure out... I couldn't think of any other way to do it than using session storage
- make put and post hooks nicer, and actually return a useful error message
- @OliverMclachlan I've included one example of rendering the error message if you want to take a look


RossStewart:
Looks like some conflicts as a couple of files are out of date. Will need to resolve these before it will let you go on.

PR #112, opened 2024-04-30
Fixing roots so that all breadcrumbs work

OliverMclachlan:
Fixing the breadcrumbs to make them work in cases that they weren't

### DIFF
--- a/src/Service.Host/client/src/components/App.js
+++ b/src/Service.Host/client/src/components/App.js
@@ -24,10 +24,7 @@ function App() {
                 <ListItem component={Link} to="/authorisation/create-group-permission">
                     <Typography color="primary">Create Group Permission</Typography>
                 </ListItem>
-                <ListItem
-                    component={Link}
-                    to="/authorisation/permission/view-individual-permission"
-                >
+                <ListItem component={Link} to="/authorisation/view-individual-permission">
                     <Typography color="primary">View Employee&apos;s Permission</Typography>
                 </ListItem>
                 <ListItem component={Link} to="/authorisation/groups">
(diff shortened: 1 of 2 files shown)


src/Service.Host/client/src/components/App.js
```diff
@@ -18,13 +18,16 @@ function App() {
...
                     <Typography color="primary">Create Individual Permission</Typography>
                 </ListItem>
-                <ListItem component={Link} to="/authorisation/permission/create/group">
+                <ListItem component={Link} to="/authorisation/create-group-permission">
                     <Typography color="primary">Create Group Permission</Typography>
                 </ListItem>
-                <ListItem component={Link} to="/authorisation/permission/view/individual">
+                <ListItem
+                    component={Link}
+                    to="/authorisation/permission/view-individual-permission"
```
lewisrenfrew, line 29: won't this still break since authorisation/permission isn't a valid route?
Could just make it authorisation/permissions for now

src/Service.Host/client/src/components/Root.js
```diff
@@ -52,17 +52,17 @@ function Root() {
...
                             <Route
                                 exact
-                                path="/authorisation/permission/create/group"
+                                path="/authorisation/create-group-permission"
                                 element={<CreateGroupPermission />}
                             />
                             <Route
                                 exact
-                                path="/authorisation/permission/view/individual"
+                                path="/authorisation/permission/view-individual-permission"
```
lewisrenfrew, line 65: same here, I don't think authorisation/permission is a valid route, so this will still render breadcrumbs with a broken 'permision' link

lewisrenfrew (approved):

PR #113, opened 2024-05-15
Add member

OliverMclachlan:
Adding the ability to add an individual member to a group


src/Facade/ResourceBuilders/MemberResourceBuilder.cs
```diff
@@ -6,6 +6,7 @@
     using Linn.Authorisation.Domain.Groups;
     using Linn.Authorisation.Resources;
     using Linn.Common.Facade;
+    using Linn.Common.Persistence;
```
lewisrenfrew, line 9: probably don't need

src/Facade/Services/IMembersFacadeService.cs
```diff
@@ -0,0 +1,12 @@
...
+    using Linn.Authorisation.Resources;
+    using Linn.Common.Facade;
+
+    public interface IMembersFacadeService
+    {
+        IResult<MemberResource> AddIndividualMember(MemberResource memberResource, string employeeUri);
+    }
+}
+    
+
```
lewisrenfrew, line 12: blank line(s)

src/Facade/Services/MembersFacadeService.cs
```diff
@@ -0,0 +1,62 @@
+﻿using Linn.Authorisation.Domain.Permissions;
+using Linn.Authorisation.Domain.Services;
+using Linn.Authorisation.Domain;
+using Linn.Common.Facade;
+using Linn.Common.Persistence;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
lewisrenfrew, line 10: usings should go inside the namespace

src/Facade/Services/MembersFacadeService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Linn.Authorisation.Facade.Services
+{
+    using Amazon.Auth.AccessControlPolicy;
```
lewisrenfrew, line 14: this one looks weird

src/Facade/Services/MembersFacadeService.cs
lewisrenfrew, line 10: also seems like quite a lot of usings - make sure they are all actually required

src/Facade/Services/MembersFacadeService.cs
```diff
@@ -0,0 +1,62 @@
...
+                return new CreatedResult<MemberResource>(new MemberResource { MemberUri = memberResource.MemberUri });
+            }
+            catch (MemberAlreadyInGroupException)
+            {
+                throw new MemberAlreadyInGroupException($"{memberResource.MemberUri} already exists in group");
+            }
+            catch (Exception ex)
+            {
+                return new BadRequestResult<MemberResource>(ex.Message);
+            }
```
lewisrenfrew, line 59: have another think about this try catch - the code in the try doesn't look like it can ever through a MemberAlreadyInGroupException. The try will need to wrap around the code that you are expecting to throw the caught exception

The catch also doesn't look right - it's just throwing the same exception again - presumably we want to handle the exception properly by returning a BadRequestResult or something

src/Service.Host/client/src/components/AddIndividualMember.js
```diff
@@ -0,0 +1,125 @@
...
+        return 0;
+    });
+
+    const handleEmployeeDropDownChange = (propertyName, newValue) => {
+        setEmployeeInput(newValue);
+    };
+
+    const handleGroupDropDownChange = (propertyName, newValue) => {
+        setGroupInput(newValue);
+    };
```
lewisrenfrew, line 63: propertyName param is unused in both these functions - probably no need to pass it?

src/Facade/Services/MembersFacadeService.cs
```diff
@@ -0,0 +1,62 @@
...
+            this.resourceBuilder = resourceBuilder;
+            this.transactionManager = transactionManager;
+            this.groupRepository = groupRepository;
+        }
+
+        public IResult<MemberResource> AddIndividualMember(MemberResource memberResource, string employeeUri)
+        {
+            var group = this.groupRepository.FindById((int)memberResource.GroupId);
+
+            var resource = this.resourceBuilder;
```
lewisrenfrew, line 42: not sure about this line - if you didn't end up using the resourceBuilder  I would just remove it from this class (or use it to build your resource as opposed to calling the MemberResource constructor on line 50

src/Service/Modules/GroupModule.cs
```diff
@@ -22,6 +23,7 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
             endpoints.MapGet("/authorisation/groups/{id:int}", this.GetGroup);
             endpoints.MapPost("/authorisation/groups", this.CreateGroup);
             endpoints.MapPut("/authorisation/groups/{id:int}", this.UpdateGroup);
+            endpoints.MapPut("/authorisation/members/{id:int}", this.AddMember);
```
lewisrenfrew, line 26: I think this should be a POST, not a PUT

src/IoC/ServiceExtensions.cs
```diff
@@ -22,9 +22,12 @@ public static IServiceCollection AddFacade(this IServiceCollection services)
             return services.AddTransient<IBuilder<Privilege>, PrivilegeResourceBuilder>()
                 .AddTransient<IBuilder<Permission>, PermissionResourceBuilder>()
                 .AddTransient<IBuilder<Group>, GroupResourceBuilder>()
-                .AddTransient<IFacadeResourceService<Privilege, int, PrivilegeResource, PrivilegeResource>, PrivilegeFacadeService>()
+                .AddTransient<IFacadeResourceService<Privilege, int, PrivilegeResource, PrivilegeResource>,
+                    PrivilegeFacadeService>()
```
lewisrenfrew, line 26: should just leave this as one line as it was before

lewisrenfrew (approved):
looks good - you could add a 'WhenPostingMember' Integration test before you merge

tests/Integration/Integration.Tests/Integration.Tests.csproj
```diff
@@ -14,6 +14,7 @@
 		<PackageReference Include="Microsoft.AspNetCore.Authentication" Version="2.2.0" />
 		<PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies" Version="2.2.0" />
 		<PackageReference Include="Microsoft.AspNetCore.TestHost" Version="7.0.9" />
+		<PackageReference Include="Microsoft.Azure.Amqp" Version="2.6.7" />
```
lewisrenfrew, line 17: not sure how this has sneaked in - can probably remove

lewisrenfrew (approved):

PR #116, opened 2024-08-01
Fixing snackbars

OliverMclachlan:


src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -40,6 +40,35 @@ public class PermissionFacadeService : IPermissionFacadeService
...
+                                       GranteeUri = ((IndividualPermission)permission).GranteeUri
+                                   });
+                }
+                else
+                {
+                    foreach (var memberUri in ((GroupPermission)permission).GranteeGroup.MemberUris())
+                    {
+                        result.Add(new PermissionResource {
+                                                                  GranteeUri = memberUri
+                                                              });
```
lewisrenfrew, line 64: formatting seems a little off here (curly brackets don't look lined up)

src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -40,6 +40,35 @@ public class PermissionFacadeService : IPermissionFacadeService
             this.groupRepository = groupRepository;
         }
 
+        public IResult<IEnumerable<PermissionResource>> GetPermissionsForPrivilege(int privilegeId)
```
lewisrenfrew, line 43: I would be tempted to make a new resource type here, something like 'EmployeeResource' that only has one field 'EmployeeUri'. I just used PermissionResource yesterday because it happened to already exist, but I think its a bit misleading here (we are not really returning a list of permissions, just a list of EmployeeUris) -  Give it a try if you feel like it, but no big deal.

src/Service/Modules/PermissionsModule.cs
```diff
@@ -23,9 +25,17 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
...
         {
-            await res.Negotiate(service.GetAllPermissionsForUser(who));
+            if (!string.IsNullOrEmpty(who))
+            {
+                await res.Negotiate(service.GetAllPermissionsForUser(who));
+            }
+            else if (privilegeId.HasValue)
+            {
+                await res.Negotiate(service.GetPermissionsForPrivilege(privilegeId.Value));
+            }
```
lewisrenfrew, line 38: I suppose these lines will be uncovered by tests - could try writing some if you are stuck for something to do next

lewisrenfrew (approved):

PR #121, opened 2024-08-26
Fixed bug

OliverMclachlan:


lewisrenfrew (changes requested):
Was this broken because there was a secondary /groups endpoint mapped to a GetApp() call or similiar? If so just remove the redundant GetApp() one (you only explicitly need to add these if the app doesn't otherwise handle requests to that endpoint)

You should also add an integration test for getting all groups - it must be missing for this bug to ever have been able to survive the travis tests and get deployed in the first place

src/Service.Host/client/src/itemTypes.js
```diff
@@ -5,7 +5,7 @@ const itemTypes = {
     employees: { url: `${config.proxyRoot}/employees?currentEmployees=true` },
     permissions: { url: `${config.appRoot}/authorisation/permissions` },
     groups: { url: `${config.appRoot}/authorisation/groups` },
-    groupData: { url: `${config.appRoot}/authorisation/groups` },
+    groupData: { url: `${config.appRoot}/authorisation/groupsInfo` },
```
lewisrenfrew, line 8: won't need this

src/Service/Modules/GroupModule.cs
```diff
@@ -19,7 +19,7 @@ public class GroupModule : IModule
     {
         public void MapEndpoints(IEndpointRouteBuilder endpoints)
         {
-            endpoints.MapGet("/authorisation/groups", this.GetAll);
+            endpoints.MapGet("/authorisation/groupsInfo", this.GetAll);
```
lewisrenfrew, line 22: i.e. this should stay as /groups- just delete the other /groups endpoint from the 'ApplicationModule'

lewisrenfrew (approved):

PR #124, opened 2025-02-28
Restrict views

OliverMclachlan:
Refactored the pages that had different pages of viewing and create

Only authorisation.admin users can create & edit groups and rivileges so we can format the name correctly

Restrict views of privileges and groups to only the admins


src/Domain/AuthorisedAction.cs
```diff
@@ -3,5 +3,21 @@
...
+
+        public const string PurchasingAdmin = "purchasing.admin";
+
+        public const string ManufacturingAdmin = "manufacturing.admin";
+
+        public const string WorkstationAdmin = "Workstation.admin";
+
+        public const string StoresAdmin = "stores.admin";
+
+        public const string SalesAdmin = "sales.admin";
```
OliverMclachlan, line 21: If there's any other you think i might need just reply to this and ill add them in

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -1,36 +1,60 @@
...
+
+    const hasViewPermission = utilities.getHref(privileges?.[0], 'view');
+    const hasCreatePermission = utilities.getHref(privileges?.[0], 'create');
 
     useEffect(() => {
-        if (data) {
-            setPrivileges(data);
+        if (privileges) {
+            const withView = privileges?.filter(privilege => utilities.getHref(privilege, 'view'));
+            setPrivilegesWithView(withView);
```
lewisrenfrew, line 45: This request itself should only ever return privileges that the user is auth'd to see - i.e. you shouldnt have to filter them on the client based on links.

lewisrenfrew (commented):
I started reviewing this PR but it started to feel a bit pointless as I went through each file making similar comments, so I'll stop that now and just summarise here:

I can see where you were going with this, i.e trying to build every privilege, permission and group resource various hypermedia links based on the currently logged in users privilges in order to control access to them, but I think doing that is either the wrong approach, or something for the future.

I say that because such links are client considerations really - niceties for REST clients to make inference about the state of the application and what operations are valid - i.e. they shouldnt be the only place domain rules are enforced like 'who is allowed to see what privileges, assign certain privileges etc.' 

Put another way, the server should never even provide information about privileges, permissions, groups etc that the user isnt authorised to see  - as opposed to what this PR is attempting, which is more along the lines of: always provide everything but with some links missing.

e.g. whether not not a client is allowed to view a resource shouldnt be decided by a 'view' link on that resource! the client should just never be provided said resource in the first instance. They should get a 401 when the requested it, or it would be missing from the list if they requested all. The fact that they would need to see the view link to decide whether they can view the resource is a bit backwards since to tell whether there was a view link or not you are already requiring them to view the resource and its links.

I think the initial work for supporting the new reqs functionality 
would start in the domain, not in resource builders. I also didnt envision any client changes would be required at all, at least initially.

As a starting point the first story should be simply 'only users with the stores.auth-manger privilege are returned privileges matching stores.* when they perform a GetAll on privileges.

First steps would be making the privileges GetAll() facade method take some information about the person making the request (i.e.their privileges) and then pass that to a domain service which works out which privileges to return.

Once you had that written and unit tested we could think about next steps

PR #125, opened 2025-03-13
Restrict view backend

OliverMclachlan:
Just split up the old pr to make it easier to review

### DIFF
--- a/src/Facade/ResourceBuilders/GroupResourceBuilder.cs
+++ b/src/Facade/ResourceBuilders/GroupResourceBuilder.cs
@@ -1,33 +1,72 @@
-﻿namespace Linn.Authorisation.Facade.ResourceBuilders
+﻿using Linn.Authorisation.Domain;
+
+namespace Linn.Authorisation.Facade.ResourceBuilders
 {
     using System.Collections.Generic;
     using System.Linq;
 
     using Linn.Authorisation.Domain.Groups;
     using Linn.Authorisation.Resources;
+    using Linn.Common.Authorisation;
     using Linn.Common.Facade;
+    using Linn.Common.Resources;
+    using Org.BouncyCastle.Utilities;
 
     public class GroupResourceBuilder : IBuilder<Group>
     {
-        private readonly MemberResourceBuilder memberResourceBuilder = new ();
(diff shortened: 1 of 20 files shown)


lewisrenfrew (changes requested):
I still think the first pass at this task should be a bit simpler than this, and shouldn't involve the resource builders. The role of a resource builder should be to construct representations of entities that have already been provided/filtered by an underlying service layer—it shouldn't be responsible for attempting to restrict access to data like this. Any code that is going to compute which data to return based on current users privileges needs to live in the domain in my opinion. 

While it's true that we do often decide which links to include on resources based on user privilege levels,  thats in cases where the user has already obtained a resource and is asking 'what other things can I do with this? what other resources are related to this?' I don't think that's at all what we're trying to achieve here - we are trying to restrict access in the first instance with some solid domain code. 

Think about it - with the approach in the PR, you're conditionally adding a 'view' link to resources that you are deciding the user is authorized to see— but those entities shouldn't ever surface from the domain layer if the user isn't auth'd to see them in the first place. 

They should never get any data and be met with an exception telling them they're not authorised to view the data they are requesting, for example.  Or in the case where they are trying to reqeust all, anything  they are not allowed to see should be filtered from their results set.

What I think we need as a first step is to augment the domain of this app with the functionality to restrict access to data based on a passed list of the clients privileges.

This should take shape as some well tested domain service methods that allow access control to the various entities (privileges, groups, permissions and such like). You could probably just focus on privileges for now and write some service methods that look something along the lines of: 

```
Privilege GetPrivilege(string name, IEnumerable<string> currrentUsersPrivileges); 
// should return the specified Privilege if the user is authorised to view it, or throw and Unauth'd exception if not
```

and
```

IEnumerable<Privilege> GetPrivileges(IEnumerable<string> currrentUsersPrivileges); 
// should return a list of only privileges the currentUser is allowed to view

```

The logic that decides whether or not a privilege is 'viewable' by a client (i.e. a currentUser) shoud live inside these sorts of methods and be well unit tested.

Once you were satisfied (all your tests were passing) you could switch the current privileges api over to use these new service methods and in turn support access control for privileges which would be a good first step/PR.

I hope all this makes sense - we can chat about this at project planning as a group if that would be helpful.

PR #128, opened 2025-04-11
Restrict groups backend

OliverMclachlan:
Merging for the restrict-views-backend was blocked by Lewis requesting changes so i merged that branch into here to do the same for groups

Only super users can access, create & edit groups for their departments (can easily change it to be only Authorisation super users can edit them)

This also restricts super users to only adding employees to groups that they have access to


src/Domain/Services/GroupService.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (userPrivileges == null || !userPrivileges.Any())
+            {
+                return new List<Group>();
+            }
+
+            var adminDepartments = userPrivileges
+                .Where(p => p.ToLower().Contains("super-user"))
+                .Select(p => p.Split('.')[0])
+                .Distinct()
+                .ToList();
```
lewisrenfrew, line 35: So the implication here is that only users with a
department.super-user 

is permitted to view groups that start with 'department.'...

i.e. we are building in the assumption that groups are always going to be department-scoped, which I'm not sure is the case (e.g. see the currently active lists of groups: https://app.linn.co.uk/authorisation/groups many aren't lexically tied to a department to allow more flexibility). 

I'm not saying this is wrong as such, just wondering if their might be a more flexible way of configuring which users are permitted to manage which groups. At least I would  change this variable name (adminDepartments) to be something a bit more correct, like superUserScopes.

As an aside, now that we are starting to actually functionally rely on the naming conventions of privileges and groups to  decide who is going to be allowed to manage them, we should probably get together and concretely agree on the conventions. Not a subject to be addressed by this PR of course, just something for us to think about.

tests/Unit/Domain.Tests/GroupServiceTests/WhenGettingAllAsAuthSuperUser.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class WhenGettingAllAsAuthSuperUser : ContextBase
+    {
+        private IEnumerable<Group> result;
+
+        [SetUp]
+        public void SetUp()
+        {
+            var groups = new List<Group>
+            {
+                new Group{Id = 1, Name = "finance.super-user"},
```
lewisrenfrew, line 23: incorrect spacing around the curly brackets (stylecop should be telling you this - double check its on and working?)

tests/Unit/Domain.Tests/GroupServiceTests/WhenGettingAllAsAuthSuperUser.cs
```diff
@@ -0,0 +1,54 @@
...
+                new Group{Id = 2, Name ="finance.do.stuuuff"},
+                new Group{ Id = 3, Name = "finance.do.hings" },
+                new Group{ Id = 4, Name = "purchasing.do.hings" }
```
lewisrenfrew, line 26: Not a huge deal but the names of these groups make them feel like privileges. I tend to like test data to feel sensible, so I'd change them to more group-y names like finance.users, purchasing.buyers or whatever to suggest they are groups of members as opposed to individual permitted operations as the current names sort of suggest

lewisrenfrew (approved):
seems ok - this is only restricting read operations as it stands (GetAll, GetById etc.). right? It probably makes sense to apply the same restrictions to write operations too (creating, updating etc.) for both privileges and groups. Not saying you should add that to this PR though - I actually think we need a show & tell discussion or similar to explore how this is going to work before you do proceed any further since, as I noted before, the work you are doing has implications for the naming conventions of privileges and groups)

PR #130, opened 2025-04-28
Group info

OliverMclachlan:
can now see what privileges a group has and all employee with a privilege including group permissions

Also put in no warning for stylecop's "Elements should be documented" and "The file header is missing or not located at the top of the file."


src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -47,23 +47,50 @@ public IResult<IEnumerable<PermissionResource>> GetPermissionsForPrivilege(int p
...
 
+            var individualUris = new List<string>();
+
             foreach (var permission in permissions)
             {
                 if (permission is IndividualPermission)
                 {
-                    result.Add(new PermissionResource { GranteeUri = ((IndividualPermission)permission).GranteeUri });
+                    result.Add(new PermissionResource{GranteeUri = ((IndividualPermission)permission).GranteeUri });
+                    individualUris.Add(((IndividualPermission)permission).GranteeUri);
```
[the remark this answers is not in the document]
lewisrenfrew: I'm not sure I really follow - what effect are the changes to this function having? Is the tracking of individualURIs just to stop the same permission being added the results list more than once? (e.g. if the Grantee has obtained the privilege both via an IndividualPermission and a GroupPermission?)

Whatever the case, you should move this logic into the domain and write good unit tests for it (then I can just read the tests to understand what is going on instead of asking dumb questions 😊)

I'm also not sure from reading the code that this function has the best name. I'd maybe change it to something like 'GetAllGranteeUrisWithPrivilege' Since all it really returns is a list of GranteeUris that point to individual users - I might even be tempted to change its return type since returning a list of PermissionResource's feels a bit wrong if the only information in each resource is a GranteeUri. Just a list of strings or something would be clearer since in my mind a valid PermissionResource should always link a privilege to a Grantee (individual or group)

src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -47,23 +47,50 @@ public IResult<IEnumerable<PermissionResource>> GetPermissionsForPrivilege(int p
...
         }
+        public IResult<IEnumerable<PermissionResource>> GetAllPermissionsForGroup(int groupId, IEnumerable<string> privileges = null)
+        {
+            var permissions = this.permissionService.GetImmediatePermissionsForGroup(groupId);
+
+            var result = new List<PermissionResource>();
+
+            foreach (var permission in permissions)
+            {
+                if (permission is GroupPermission)
```
lewisrenfrew, line 82: isn't this always true? It looks like permissionService.GetImmediatePermissionsForGroup can only ever return GroupPermissions anyways. I think you can remove this little piece of logic and then this function can just be a simple pure facade service function that does nothing but request data from the domain and then translate it to an IResult. As a general rule this is all we really want facade services to be doing - when you start to find yourself writing checks, if statements and such like then thats a good sign that the code should be moved to the domain

src/Persistence/PermissionsRepository.cs
```diff
@@ -46,8 +46,9 @@ public Permission FindBy(Expression<Func<Permission, bool>> expression)
         public IQueryable<Permission> FilterBy(Expression<Func<Permission, bool>> expression)
         {
             var indivPermissions = this.serviceDbContext.Permissions.Where(p => p is IndividualPermission).Include(x => x.Privilege).ToList();
-            var groupPermissions = this.serviceDbContext.Permissions.Where(p => p is GroupPermission).Include(x => ((GroupPermission)x).GranteeGroup).Include(x => x.Privilege).ToList();
-            var allPermissions = groupPermissions.Concat(indivPermissions).AsQueryable().Where(expression);
+            var groupPermissions = this.serviceDbContext.Permissions.Where(p => p is GroupPermission).Include(p => ((GroupPermission)p).GranteeGroup)
+                .ThenInclude(g => g.Members).Include(p => p.Privilege).ToList();
```
lewisrenfrew, line 50: is the only difference here that we are now .ThenInclude()'ing every Groups members? Why do we need to pull back that information from here in the PermissionRepository?

src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -47,23 +47,50 @@ public IResult<IEnumerable<PermissionResource>> GetPermissionsForPrivilege(int p
...
             var result = new List<PermissionResource>();
 
+            var individualUris = new List<string>();
+
             foreach (var permission in permissions)
             {
                 if (permission is IndividualPermission)
                 {
-                    result.Add(new PermissionResource { GranteeUri = ((IndividualPermission)permission).GranteeUri });
+                    result.Add(new PermissionResource{GranteeUri = ((IndividualPermission)permission).GranteeUri });
```
lewisrenfrew, line 56: some invalid spacings around { here

src/Service/Modules/PermissionsModule.cs
```diff
@@ -33,6 +34,14 @@ private async Task GetAllPermissionsForUser(
             }
         }
 
+        private async Task GetAllPermissionsForGroup(
+            HttpResponse res,
+            int groupId,
+            IPermissionFacadeService service)
+        { 
+            await res.Negotiate(service.GetAllPermissionsForGroup(groupId));
+        }
```
lewisrenfrew, line 43: Admittedly without having thought about this too much, I didn't expect we'd need a new API endpoint to achieve this.

I wonder if there is some way of just building all the permissions the group has onto the resource that is returned when we request the group, i.e. usiing the /authorisation/permissions/{groupId:int} endpoint

OliverMclachlan: yea could probably do something using the permission Repository with the group id

src/Persistence/PermissionsRepository.cs
OliverMclachlan: means that group members weren't included and they weren't getting displayed despite having the permission

src/Domain.LinnApps/Domain.LinnApps.csproj
```diff
@@ -4,6 +4,8 @@
 		<TargetFramework>net7.0</TargetFramework>
 		<AssemblyName>Linn.Authorisation.Domain.LinnApps</AssemblyName>
 		<RootNamespace>Linn.Authorisation.Domain.LinnApps</RootNamespace>
+		<NoWarn>$(NoWarn);SA1600</NoWarn>
```
richardiphillips, line 7: you should be able to set resharper / stylecop settings for you whole computer rather than having to have it in a particular solution

src/Domain/Services/IPermissionService.cs
```diff
@@ -11,5 +11,6 @@ public interface IPermissionService
         IEnumerable<Permission> GetAllPermissionsForPrivilege(int privilegeId);
 
         IEnumerable<Permission> GetAllPermissionsForUser(string who);
+        List<string> GetAllGranteeUris(IEnumerable<Permission> permissions);
```
richardiphillips, line 14: stylecop would say there should be a blank line i reckon. 

Also you almost certainly don't need to pass a List out. Either IEnumerable or IList will use less resource

src/Facade/ResourceBuilders/PermissionResourceBuilder.cs
```diff
@@ -20,7 +20,7 @@ public object Build(Permission model, IEnumerable<string> claims)
                     Privilege = model.Privilege.Name,
                     PrivilegeId = model.Privilege.Id,
                     Links = this.BuildLinks(model).ToArray(),
-                    Id = model.Id
+                    Id = model.Id,
```
richardiphillips, line 23: hanging commas

src/Facade/ResourceBuilders/PermissionResourceBuilder.cs
```diff
@@ -31,7 +31,7 @@ public object Build(Permission model, IEnumerable<string> claims)
                 Links = this.BuildLinks(model).ToArray(),
                 Id = model.Id,
                 GranteeGroupId = ((GroupPermission)model).GranteeGroup.Id,
-                GroupName = ((GroupPermission)model).GranteeGroup.Name
+                GroupName = ((GroupPermission)model).GranteeGroup.Name,
```
richardiphillips, line 34: hanging commas

src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -41,25 +41,29 @@ public PermissionFacadeService(
             this.groupRepository = groupRepository;
         }
 
-        public IResult<IEnumerable<PermissionResource>> GetPermissionsForPrivilege(int privilegeId, IEnumerable<string> privileges = null)
+        public IResult<List<string>> GetPermissionsForPrivilege(int privilegeId, IEnumerable<string> privileges = null)
```
richardiphillips, line 44: again don't need full fat List

tests/Unit/Domain.Tests/PermissionServiceTests/WhenGettingAllUris.cs
```diff
@@ -0,0 +1,80 @@
...
+        [Test]
+        public void ShouldCallPermissionsRepository()
+        {
+            this.result.Should().Contain("/employees/1");
+            this.result.Should().Contain("/employees/2");
+            this.result.Should().Contain("/employees/3");
+            this.result.Should().Contain("/employees/4");
+        }
+    }
+}
```
richardiphillips, line 80: github likes a carriage return at the end.

tests/Integration/Integration.Tests/PermissionsModuleTests/WhenGettingPermissionsForGroup.cs
```diff
@@ -0,0 +1,78 @@
...
+                    GranteeGroup = this.group,
+                    Privilege = new Privilege { Id = 1, Name = "Privilege 1" }
+                },
+                new GroupPermission
+                {
+                    GranteeGroup = this.group,
+                    Privilege = new Privilege { Id = 2, Name = "Privilege 2" }
+                }
+            };
+
```
richardiphillips, line 47: blank

richardiphillips (approved):
mainly stylecop. i'm assuming Lewis will check the actual code!

src/Facade/ResourceBuilders/PermissionResourceBuilder.cs
OliverMclachlan: Its a stylecop thing i'll add that to the list of things to disable

tests/Unit/Domain.Tests/PermissionServiceTests/WhenGettingAllUris.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Linn.Authorisation.Domain.Tests.PermissionServiceTests
+﻿using System.Collections;
```
richardiphillips, line 1: check new stylecop is working here

tests/Unit/Domain.Tests/PermissionServiceTests/WhenGettingAllUris.cs
```diff
@@ -54,12 +56,15 @@ public void SetUp()
...
-                                  };
+            {
+                new IndividualPermission("/employees/1", this.privilege, "/employees/7004"),
+                new IndividualPermission("/employees/2", this.privilege, "/employees/7004"),
+                new GroupPermission(this.group, this.privilege, "/employees/7004")
+            };
+
             this.result = this.Sut.GetAllGranteeUris(permissions);
+
+
```
richardiphillips, line 67: blanks

src/Facade/ResourceBuilders/GroupResourceBuilder.cs
```diff
@@ -9,20 +9,25 @@
 
     public class GroupResourceBuilder : IBuilder<Group>
     {
-        private readonly MemberResourceBuilder memberResourceBuilder = new();
+        private readonly MemberResourceBuilder memberResourceBuilder = new ();
+
+        private readonly PermissionResourceBuilder permissionResourceBuilder = new ();
```
lewisrenfrew, line 14: can probably use dependency injection here? Can't think of any reason why not to off the top of my head

src/Facade/Services/PermissionFacadeService.cs
```diff
@@ -41,25 +41,29 @@ public PermissionFacadeService(
...
-                    foreach (var memberUri in ((GroupPermission)permission).GranteeGroup.MemberUris())
+                result.Add(new PermissionResource
                     {
-                        result.Add(new PermissionResource { GranteeUri = memberUri });
-                    }
-                }
+                        GroupName = ((GroupPermission)permission).GranteeGroup.Name,
+                        PrivilegeId = permission.Privilege.Id,
+                        Privilege = permission.Privilege.Name,
+                    });
```
lewisrenfrew, line 66: I prefer LINQ .Select for this sort of thing, i.e. 

var result = permissions.Select(p => new PermissionResource { blah })

src/Resources/GroupResource.cs
```diff
@@ -12,5 +12,7 @@ public class GroupResource
         public int Id { get; set; }
 
         public IEnumerable<MemberResource> Members { get; set; }
+
+        public IEnumerable<PermissionResource> Permission { get; set; }
```
lewisrenfrew, line 16: Permissions plural is a better field name

src/Service.Host/client/src/components/Group.js
```diff
@@ -152,12 +163,52 @@ function Group({ creating }) {
...
+
+                            <Grid item xs={12}>
+                                <DataGrid
+                                    rows={
+                                        groupMembers?.map(e => ({
+                                            ...e,
+                                            id: e?.id
+                                        })) || []
+                                    }
+                                    columns={employeeColumns}
```
lewisrenfrew, line 181: in theory group members can themselves be other groups, but I know we've never actually created that scenario yet. Just a note that this UI might need to change if a group has a member that isn't an 'employee' but another group

src/Service/Modules/PermissionsModule.cs
```diff
@@ -17,6 +17,7 @@ public class PermissionsModule : IModule
         public void MapEndpoints(IEndpointRouteBuilder endpoints)
         {
             endpoints.MapGet("/authorisation/permissions", this.GetAllPermissionsForUser);
+            endpoints.MapGet("/authorisation/permissions/{groupId:int}", this.GetAllPermissionsForGroup);
```
lewisrenfrew, line 20: Thinking about a clean and clear permissions API url scheme... this endpoint (to filter permissions based on a group parameter) seems a little inconsistent with the rest of the endpoints in this module in a couple of ways

1) the precedent set by the delete url is that the location of the permission to be deleted is specified by the following: /authorisation/permissions/id-of-permission
so I think for good API design (clarity) we should stick to that convention, i.e. permissions are uniquely identified as /authorisations/permissions/id-of-permission
that means that this URL to GetAllPermissionsForGroup conflicts with that convention (i would naively assume that this url is going to fetch me the permission with id: groupId which is wrong) - I hope that makese sense.

2) other similar filtering endpoints in this module, i.e. GetPermissionsForPrivilege and GetAllPermissionsForUser follow a scheme where the filter parameter seems to be a query parameter as opposed to an id directly encoded in the url as you have in this GetAllPermissionsForGroup case - I would stick to one approach for consistency, and I personally prefer query parameters for filtering operations like this

you could even explroe a single /permissions collection endpoint that support multiple different query parameters? i.e. if the client passes a ?groupId=123 then apply group filtering, if they pass ?privlegeId=456 then apply permission filtering, etc.

lewisrenfrew (commented):
I was in the process of leaving these comments before you merged, so just something to think about for your next PR

PR #131, opened 2025-05-01
Delete member

OliverMclachlan:


src/Facade/Services/IGroupFacadeService.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using System.Collections.Generic;
+
+    using Linn.Authorisation.Domain.Groups;
+    using Linn.Authorisation.Resources;
+    using Linn.Common.Facade;
+
+    public interface IGroupFacadeService : IFacadeResourceService<Group, int, GroupResource, GroupResource>
+    {
+        void DeleteMember(int groupId, string employeeUri, int memberId, IEnumerable<string> userPrivileges = null);
```
lewisrenfrew, line 11: Why do we need this method on the GroupFacadeService if we've added similar to an IMembersFacadeService? 
It looks like its the IMembersFacadeService.DeleteMember you are calling from the module.. so is this one redundant?

src/Facade/Services/MembersFacadeService.cs
```diff
@@ -42,5 +46,18 @@ public IResult<MemberResource> AddIndividualMember(MemberResource memberResource
...
+
+        public IResult<MemberResource> DeleteMember(int memberId)
+        {
+            var member = this.memberRepository.FindById(memberId);
+
+            this.memberRepository.Remove(member);
+
+            this.transactionManager.Commit();
+
+            var result = new MemberResource { GroupId = null, MemberUri = null };
```
lewisrenfrew, line 58: I prefer returning the resource representation of the deleted object for DELETE requests

src/Persistence/MemberRepository.cs
```diff
@@ -0,0 +1,49 @@
+﻿namespace Linn.Authorisation.Persistence
+{
+    using System;
+    using System.Linq;
+    using System.Linq.Expressions;
+
+    using Linn.Authorisation.Domain.Groups;
+    using Linn.Common.Persistence;
+
+    public class MemberRepository : IRepository<Member, int>
```
lewisrenfrew, line 10: can we not just use the generic EntityFrameworkRepository here? None of the overrides below  seem  to be doing anything different to the  common code (apart from ordering, which arguably should be specified by consuming code anyway)

lewisrenfrew (approved):

PR #132, opened 2025-05-08
Change super user to auth manager

OliverMclachlan:
Just doing this in a seperate pr so the other branch doesn't have as many files changed


src/Domain/Services/GroupService.cs
```diff
@@ -18,7 +18,7 @@ public GroupService(IRepository<Group, int> groupRepository)
 
         public IEnumerable<Group> GetAllGroupsForUser(IEnumerable<string> userPrivileges = null)
         {
-            if (userPrivileges.Contains("authorisation.super-user"))
+            if (userPrivileges.Contains("authorisation.auth-manager"))
```
lewisrenfrew, line 21: I would use the const here (the one in the AuthorisedAction file above) just to mitigate against the possibility of future changes  / typos

src/Domain/Services/GroupService.cs
```diff
@@ -50,7 +50,7 @@ public IEnumerable<Group> GetAllGroupsForUser(IEnumerable<string> userPrivileges
 
         public Group GetGroupById(int groupId, IEnumerable<string> userPrivileges = null)
         {
-            if (userPrivileges.Contains("authorisation.super-user"))
+            if (userPrivileges.Contains("authorisation.auth-manager"))
```
lewisrenfrew, line 53: and here

src/Domain/Services/PrivilegeService.cs
```diff
@@ -49,7 +49,7 @@ public IEnumerable<Privilege> GetAllPrivilegesForUser(IEnumerable<string> userPr
 
         public Privilege GetPrivilegeById(int privilegeId, IEnumerable<string> userPrivileges = null)
         {
-            if (userPrivileges.Contains("authorisation.super-user"))
+            if (userPrivileges.Contains("authorisation.auth-manager"))
```
lewisrenfrew, line 52: and here

lewisrenfrew (approved):

PR #135, opened 2025-05-26
Show and tell refactoring

OliverMclachlan:
Adding log tables, more checks, menu, etc from show and tell


src/Service.Host/client/src/components/Groups.js
```diff
@@ -14,36 +13,26 @@ import Page from './Page';
...
+        const fa = a.name?.toLowerCase() || '';
+        const fb = b.name?.toLowerCase() || '';
+        return fa.localeCompare(fb);
+    }
 
-    groups.sort((a, b) => {
-        const fa = a.name.toLowerCase();
-        const fb = b.name.toLowerCase();
+    const sortedGroupsInfo = [...(data || [])]?.sort(sorting);
+    const sortedActiveGroupsInfo = [...(activePrivileges || [])]?.sort(sorting);
```
lewisrenfrew, line 29: as above

src/Service.Host/client/src/components/Groups.js
```diff
@@ -70,9 +59,33 @@ function Groups() {
...
+                            Active Only
+                        </Typography>
+                        <DataGrid
+                            rows={activeOnly ? sortedActiveGroupsInfo : sortedGroupsInfo}
+                            getRowId={row => row?.id}
+                            columns={columns}
+                            editMode="cell"
+                            onRowClick={clicked => {
+                                navigate(`/authorisation/groups/${clicked.row.id}`);
+                            }}
```
lewisrenfrew, line 80: do groups have a rel: self link? If so make use of it here

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -14,57 +12,64 @@ import Page from './Page';
...
+        const fa = a.name?.toLowerCase() || '';
+        const fb = b.name?.toLowerCase() || '';
+        return fa.localeCompare(fb);
+    }
 
-    privileges?.sort((a, b) => {
-        const fa = a?.name.toLowerCase();
-        const fb = b?.name.toLowerCase();
+    const sortedPrivilegesInfo = [...(data || [])]?.sort(sorting);
+    const sortedActivePrivilegesInfo = [...(activePrivileges || [])]?.sort(sorting);
```
lewisrenfrew, line 28: can be more readable like the other examples

src/Service.Host/client/src/components/Privileges.js
```diff
@@ -14,57 +12,64 @@ import Page from './Page';
...
+                            Active Only
+                        </Typography>
+
+                        <DataGrid
+                            rows={activeOnly ? sortedActivePrivilegesInfo : sortedPrivilegesInfo}
+                            getRowId={row => row?.id}
+                            columns={columns}
+                            editMode="cell"
+                            onRowClick={clicked => {
+                                navigate(`/authorisation/privileges/${clicked.row.id}`);
```
lewisrenfrew, line 63: again, rel:self link would be nice

src/Service.Host/client/src/components/ViewEmployeesPermission.js
```diff
@@ -0,0 +1,165 @@
...
+                </Grid>
+
+                {permissions && (
+                    <Grid item xs={12}>
+                        <DataGrid
+                            rows={
+                                permissions?.map(e => ({
+                                    ...e,
+                                    id: e?.id
+                                })) || []
```
lewisrenfrew, line 144: do we need this map here? looks like we are just renaming the id field to be 'id', which is redundant

src/Service.Host/client/src/components/ViewPermissionUsers.js
```diff
@@ -80,8 +151,29 @@ function ViewPermissionUsers() {
...
             </Grid>
 
+            {permissions && (
+                <Grid item xs={12}>
+                    <DataGrid
+                        rows={
+                            permissions?.map(e => ({
+                                ...e,
+                                id: e?.id
+                            })) || []
```
lewisrenfrew, line 161: redundant map again?

tests/Integration/Integration.Tests/PermissionsModuleTests/WhenGettingAllPermissionForAPrivilege.cs
```diff
@@ -0,0 +1,112 @@
...
+                                                                new IndividualMember
+                                                                    {
+                                                                        Id = 6,
+                                                                        MemberUri = "/employees/5678"
+                                                                    }
+                                                            }
+                                                    }
+                                            }
+                                    }
+            };
```
lewisrenfrew, line 60: it doesnt look like this members list has any effect on the outcome of the test, so id just delete in the name of keeping test data as simple as is sufficient

src/Service.Host/client/src/tests/Groups.test.js
```diff
@@ -28,19 +29,20 @@ describe('When groups ', () => {
...
-        expect(listItems[2].href).toContain('authorisation/groups/4');
-        expect(listItems[3].href).toContain('authorisation/groups/3');
-        expect(listItems[4].href).toContain('authorisation/groups/2');
+        await waitFor(() => {
+            const cells = getAllByRole('cell');
+            const GroupNames = cells
+                .filter(cell => cell.getAttribute('data-field') === 'name')
+                .map(cell => cell.textContent);
+            expect(GroupNames).toEqual(['a.group', 'b.group', 'd.group']);
+        });
```
lewisrenfrew, line 45: i don't expect we need the async/await and waitFor here, or at least I can't think of a reason why we would...

lewisrenfrew (approved):
just some comments r.e. using API results to update UI state

